PR #4, opened 2017-11-10
Infrastructure

### LINKED ISSUE #2: Enable testing infrastructure
- [x] Docker container with dealII
- [x] Jenkins project
- [x] Github hooks

### LINKED ISSUE #3: Setup clang-formatting for code

Rombur:
close #2 
close #3


CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 3.9)
+
+PROJECT(mfmg LANGUAGES CXX VERSION ${MFMG_VERSION})
+
+SET(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+INCLUDE(SetupTPLs)
+INCLUDE(SetupMFMG)
+
+IF(${MFMG_ENABLE_COVERAGE})
+  INCLUDe(CodeCoverage)
```
aprokop: INCLUDe -> INCLUDE :)

CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+IF(${MFMG_ENABLE_ClangFormat})
+  INCLUDE(CodeFormat)
+ENDIF()
+
+IF(${MFMG_ENABLE_DOCUMENTATION})
+  ADD_SUBDIRECTORY(doc)
+ENDIF()
+
+INCLUDE_DIRECTORIES(include)
+FILE(GLOB MFMG_HEADERS include/*.hh)
```
aprokop: I would prefer headers to be `hpp` and not `hh`

ci/Dockerfile
```diff
@@ -0,0 +1,189 @@
...
+    cd ${P4EST_SOURCE_DIR} && \
+    wget --quiet https://www.dealii.org/8.5.0/external-libs/p4est-setup.sh && \
+    bash ./p4est-setup.sh ${P4EST_ARCHIVE} ${P4EST_INSTALL_DIR} && \
+    rm -rf ${P4EST_ARCHIVE} && \
+    rm -rf ${P4EST_BUILD_DIR} && \
+    rm -rf ${P4EST_SOURCE_DIR}
+ENV P4EST_DIR=${INSTALL_DIR}/p4est
+
+# Install deal.II
+RUN export DEAL_II_URL=https://github.com/dealii/dealii.git && \
```
aprokop: Is it a good idea to run on the tip of deallii?

doc/conf.py
```diff
@@ -0,0 +1,171 @@
...
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'mfmg'
+copyright = u'2017, Bruno Turcksin, Andrey Prokopenko'
```
aprokop: No Wayne?

include/mfmg/dummy.hh
```diff
@@ -0,0 +1,18 @@
+/*************************************************************************
+ * Copyright (c) 2017 by the mfmg authors                                *
+ * All rights reserved.                                                  *
+ *                                                                       *
+ * This file is part of the mfmg libary. mfmg is distributed under a BSD *
+ * 3-clause license. For the licensing terms see the LICENSE file in the *
+ * top-level directory                                                   *
+ *************************************************************************/
```
aprokop: I recently read https://reuse.software/practices/. May not be a bad idea to follow that.

ci/Dockerfile
Rombur: That will be necessary for CUDA. CUDA support has not been release yet. I could use a hash if you think it makes sense

doc/conf.py
Rombur: Forgot. Thanks.

include/mfmg/dummy.hh
Rombur: I don't see what would be different with what we are doing now. Can you explain?

ci/Dockerfile
aprokop: Yes, I think we should freeze the version we test against. Otherwise, we are susceptible to dealii breakage.

include/mfmg/dummy.hh
aprokop: It's not different in the spirit, just in details as the standardize the language and reference SPDX which allows for automated tracking of licenses.

PR #6, opened 2017-11-15
Laplace 

Rombur:
Add a simple solver for the Laplace. This can be used to test our preconditioner.


Rombur:
Alright the PR passes the test now.

tests/laplace.hpp
```diff
@@ -0,0 +1,240 @@
...
+
+  _dof_handler.distribute_dofs(_fe);
+
+  // Get the IndexSets
+  _locally_owned_dofs = _dof_handler.locally_owned_dofs();
+  dealii::DoFTools::extract_locally_relevant_dofs(_dof_handler,
+                                                  _locally_relevant_dofs);
+
+  // Resize the vectors
+  _locally_relevant_solution.reinit(_locally_owned_dofs, _locally_relevant_dofs,
```
aprokop: I'm not sure I understand why it takes 2 "maps" in.

Rombur: This is the only way to create a ghosted vector in deal.II. It allows you to know what is ghosted and what is not. I don't think this is necessary with Trilinos vectors but PETSc vectors behave very differently and we want to have only one interface to switch between all the type of vectors that are supported http://dealii.org/8.5.0/doxygen/deal.II/DEALGlossary.html#GlossGhostedVector

Rombur:
Done

PR #9, opened 2017-11-16
Compile boost with rpath support

Rombur:
By default, Boost does not support rpath. This is a problem if a boost library
depends on another one (eg. boost.mpi depends on boost.serialization) and boost
lib is not in the LD_LIBRARY_PATH.


Rombur:
Ready for merge. I will rebase #6 once this has been merged.

PR #11, opened 2017-11-30
Modify the license so it is recognized by github

Rombur:


aprokop:
I'm not sure that will help. I have the exact same problem with ForTrilinos where I added the license *after* creation of the project. And it's not recognized despite the default text. You can try running `licensee` and see if it picks it up. For me, it produced NCSA 64% similarity.

Rombur:
Let's try this. I get 96% similarity with BSD. I had 64% before I remove the italic text.

aprokop:
By the way, you need to remove `(c)`. I was told it means that the product was registered with U.S. copyright office. I think ORNL's licenses were updated the same way for open-source.

Rombur:
Darn, it won't work. The threshold is at 98%.

PR #13, opened 2017-12-08
Add Arpack to the Docker image.

Rombur:


Rombur:
Don't merge this. There is problem with jenkins when we use merge on master. I want to fix it first.

Rombur:
OK it should be fixed

PR #15, opened 2017-12-19
Potpourri

aprokop:


.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
Rombur: Not a fan of adding this. You want to ignore it because you build the code there not because it is a file that should be ignored. Can you put that in your `.git/info/exclude` instead?

cmake/SetupMFMG.cmake
```diff
@@ -12,12 +12,12 @@ set(CMAKE_CXX_FLAGS_DEBUG "-g")
 
 # Disable boost asserts in Release mode and set MFMG_DEBUG in Debug mode
 MESSAGE("-- Build type: ${CMAKE_BUILD_TYPE}")
-IF(CMAKE_BUILD_TYPE MATCHES "Release")
+IF(CMAKE_BUILD_TYPE MATCHES "Release" OR CMAKE_BUILD_TYPE MATCHES "RelWithDebInfo")
```
Rombur: I think that for `RelWithDebInfo` we should set both `-O3` and `-g` flags. Otherwise it is not in a release build and there is no debug info.

.gitignore
aprokop: OK.

cmake/SetupMFMG.cmake
aprokop: CMake does it for you.

Rombur: OK great.

aprokop:
@Rombur Updated the PR:
- Removed .gitignore
- Added git hooks for clang-format (similarly to DTK)

PR #17, opened 2017-12-22
Cuda cmake (ready for review but do NOT merge)

Rombur:
This PR is can be reviewed but do not merge it yet. I want to drop a commit which adds a dummy test. 

This PR removes the dependency with boost::mpi because it does not compile with nvcc. Also by default we now only produce a static library. Build for a shared library only works if cuda is disabled.


aprokop:
@Rombur Do you want to clean the patches, or should I just squash?

Rombur:
I want to drop a commit. I'll do it and merge.

PR #18, opened 2018-01-04
Cuda sparse matrix

Rombur:
This PR does mainly two things: 
 1. Add some asserts and exceptions
 2. Add two functions to move sparse matrices to the gpu


Rombur:
_ping_

include/mfmg/exceptions.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+inline void ASSERT_THROW_NOT_IMPLEMENTED()
+{
+  NotImplementedExc exception;
+  throw exception;
+}
+
+#if MFMG_WITH_CUDA
+#if MFMG_DEBUG
+inline void ASSERT_CUDA(cudaError_t error_code)
```
aprokop: What about providing macro 
```
ASSERT_CUDA(cuda_code) \
{ \
  auto error_code = cuda_code; \
  auto message = cudaGet...; \
  ASSERT(error_code == cudaSuccess, ...) \
}
```

include/mfmg/utils.cuh
```diff
@@ -0,0 +1,146 @@
...
+            std::vector<int> const &column_index,
+            std::vector<int> const &row_ptr)
+{
+  unsigned int const nnz = val.size();
+  unsigned int const row_ptr_size = row_ptr.size();
+
+  // Copy the elements to the gpu
+  ScalarType *val_dev;
+  cudaError_t error_code = cudaMalloc(&val_dev, nnz * sizeof(ScalarType));
+  ASSERT_CUDA(error_code);
```
aprokop: See comment above if we could write
```
ASSERT_CUDA(cudaMalloc(&val_dev, nnz * sizeof(ScalarType)));
```

include/mfmg/utils.cuh
```diff
@@ -0,0 +1,146 @@
...
+  error_code = cudaMemcpy(val_dev, &val[0], nnz * sizeof(ScalarType),
```
aprokop: You need to check that sizes are not 0.

include/mfmg/utils.cuh
```diff
@@ -0,0 +1,146 @@
...
+                          cudaMemcpyHostToDevice);
+  ASSERT_CUDA(error_code);
+
+  // Copy the column indices to the gpu
+  int *column_index_dev;
+  error_code = cudaMalloc(&column_index_dev, nnz * sizeof(int));
+  ASSERT_CUDA(error_code);
+  error_code = cudaMemcpy(column_index_dev, &column_index[0], nnz * sizeof(int),
```
aprokop: check sizes.

include/mfmg/utils.cuh
```diff
@@ -0,0 +1,146 @@
...
+
+#include <tuple>
+
+namespace mfmg
+{
+namespace internal
+{
+template <typename ScalarType>
+std::tuple<ScalarType *, int *, int *>
+copy_to_gpu(std::vector<ScalarType> const &val,
```
aprokop: Maybe replace these function by a simpler one?
```
template<typename Type>
Type* copy_to_gpu(std::vector<Type> const& vec) {
}
```
These would reduce copy-paste code.

include/mfmg/utils.cuh
```diff
@@ -0,0 +1,146 @@
...
+  return std::make_tuple(val_dev, column_index_dev, row_ptr_dev);
+}
+}
+
+/**
+ * Convert a dealii::SparseMatrix to the regular CSR format and move the data to
+ * the GPU.
+ */
+template <typename ScalarType>
+std::tuple<ScalarType *, int *, int *>
```
aprokop: How about creating a struct or something similar and using that instead of tuple? That would make it clear that it is a device matrix.

include/mfmg/utils.cuh
```diff
@@ -0,0 +1,146 @@
...
+ * Move a dealii::TrilinosWrappers::SparseMatrix to the GPU.
+ */
+std::tuple<double *, int *, int *>
+convert_matrix(dealii::TrilinosWrappers::SparseMatrix const &sparse_matrix)
+{
+  unsigned int const n_local_rows = sparse_matrix.local_size();
+  std::vector<double> val;
+  std::vector<int> column_index;
+  std::vector<int> row_ptr(n_local_rows + 1);
+  for (unsigned int row = 0; row < n_local_rows; ++row)
```
aprokop: Isn't there a function to directly get all three arrays out of a Trilinos matrix? There is definitely such a function in Epetra.

include/mfmg/exceptions.hpp
```diff
@@ -0,0 +1,67 @@
...
+#define EXCEPTIONS_HPP
+
+#include <exception>
+#include <stdexcept>
+#include <string>
+
+namespace mfmg
+{
+#if MFMG_DEBUG
+inline void ASSERT(bool cond, std::string const &message)
```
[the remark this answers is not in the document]
Rombur: At first I tried to use a macro but the message was never printed correctly. However I have another branch that depends on this where I use stacktrace from boost and it gives you a full stracktrace. So it is even better than a macro see https://github.com/Rombur/mfmg/commit/bf4210b44b69358a89b72ae7f8b7d16121762c4f The only problem is that it assumes that `dl` and `backtrace` are in your path but they should always be on Linux.

include/mfmg/utils.cuh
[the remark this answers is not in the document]
Rombur: I have used the opposite ordering every where :disappointed:

include/mfmg/exceptions.hpp
Rombur: I know some people do that but I don't like putting code that needs to be executed inside a macro

include/mfmg/utils.cuh
Rombur: Why would I want that? I still need to reorder all the column indices.

include/mfmg/utils.cuh
Rombur: I am not sure what you mean. There is only one function `copy_to_gpu`.

include/mfmg/utils.cuh
Rombur: OK I'll change it.

include/mfmg/utils.cuh
aprokop: In your `copy_to_gpu` you provide three arrays of different types, and return a tuple of pointers. I propose modifying it that instead it takes in a single array, and being called three times.

include/mfmg/utils.cuh
aprokop: You don't reorder for Trilinos from what I see.

Rombur: Actually `column_index` is not the same as the underlying `Tpetra` array that's why I need to call `global_column_index`. This function just call `sparse_matrix.GCID(indices[i])` ( or `sparse_matrix.GCID64(indices[i])`)

Rombur:
I have encapsulate the pointers in a struct

aprokop:
Is this now ready for the second look?

Rombur:
yeah the only thing I didn't change is to get the three csr vectors at once from Trilinos because the datatype is different than the one I want

Rombur:
_ping_

aprokop:
@Rombur Looks fine to my untrained eye, but I would still like you to check the sizes in `copy_to_gpu` so that `&val[0]` does not segfault.

Rombur:
There is an assert that checks that the size is positive https://github.com/ORNL-CEES/mfmg/pull/18/files#diff-6f8744281bd3500378d3c5a71a3981b2R34

PR #21, opened 2018-01-12
Split AMGe into AMGe and AMGe_host

Rombur:
In order to prepare the code for CUDA, we create a base class AMGe and move CPU
only code to AMGe_host. 

This PR moves code around and improve the documentation. There is no change in functionality.


Rombur:
_ping_

PR #23, opened 2018-01-31
Compute eigenvector on the device

Rombur:
This PR:
 * add assert for cuSPARSE and add support for backtrace in the assert.
 * compute the eigenvalues and the eigenvectors on the device. For now, we use cuSolver and we compute all the eigenvectors and all the eigenvalues. We will need to implement our own batched Lanczos algorithm later
 * add a test


Rombur:
I don't know why but on fetnat I need to add `-fopenmp` for `cuSolver` to work. I don't need it on my own machine...

Rombur:
Fixed. `cuSolver` requires `OpenMP`, I don't know why it worked in my machine without the flag.

Rombur:
This has been sitting here for a while. Since there are no comment, I assume this can be merged.

PR #28, opened 2018-02-20
Misc change

Rombur:
A bunch of commits necessary to get the two-grid algorithm working.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright 2017 UT-Battelle, LLC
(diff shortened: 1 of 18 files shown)


aprokop:
I don't think I like putting matrix-matrix multiplication into the `AMGe_host` setup. I understand that you need this to make 2-grid hierarchy working. I can merge it in this form, however, I'll move out the matrix-matrix multiplication stuff out of `setup` and let `Hierarchy` handle it. The `setup` should really be replaced by a single `build_restrictor` function.

Rombur:
Yeah that's fine. I don't have a hierarchy so I'v put everything in the setup.

PR #29, opened 2018-02-27
Misc changes

aprokop:
Interface independent changes


include/mfmg/exceptions.hpp
```diff
@@ -77,6 +77,7 @@ class NotImplementedExc : public std::exception
 
 inline void ASSERT_THROW_NOT_IMPLEMENTED()
 {
+  std::cerr << boost::stacktrace::stacktrace();
```
Rombur: If you want this you should guard it with `MFMG_WITH_STACKTRACE` otherwise the code won't compile when we don't have stacktrace.

source/utils.cc
```diff
@@ -0,0 +1,44 @@
...
+ * All rights reserved.                                                  *
+ *                                                                       *
+ * This file is part of the mfmg libary. mfmg is distributed under a BSD *
+ * 3-clause license. For the licensing terms see the LICENSE file in the *
+ * top-level directory                                                   *
+ *                                                                       *
+ * SPDX-License-Identifier: BSD-3-Clause                                 *
+ *************************************************************************/
+
+#include <mfmg/exceptions.hpp>
```
Rombur: So the PR does not compile because of this. The problem is that `exceptions.hpp` contains cuda code but `utils.cc` is sent to `gcc` not `nvcc`. The solution is to split `exceptions.hpp` into `excceptions.hpp` and `exceptions.cuh`. `exceptions.cuh` includes `exceptions.hpp` and contains all the cuda code.

aprokop: Hmm, I tried to work around by putting the exceptions CUDA code inside `#ifdef __CUDACC__` or `#ifdef __CUDA_ARCH__` but that leads to CUDA tests missing `cusparseHandle_t` that I cannot reenable. Something's weird here. I would really prefer not to split the `exceptions.hpp` files so that we won't have to think about it.

aprokop: OK, fixed it. My proposed solution (using `__CUDACC__`) actually works, but I was accidentally including `cusparse.h' *inside* `mfmg` namespace. Bummer.

PR #32, opened 2018-03-08
Better looking convergence rate ;)

### LINKED ISSUE #31: Reproduce 2003 spectral amge paper example
```
In particular, we consider the Poisson equation discretized with bilinear functions
on a 32 × 32 square grid of square elements. Homogeneous Dirichlet boundary conditions
are assumed, and they are eliminated from the matrix during discretization.
Standard 2 × 2 agglomeration is forced on each level.
```

**Unstaggered elements**

| metric              |       |
| ------------------- | ----- |
| number of levels    | 2     |
| convergence rate    | 0.04  |
| grid complexity     | 1.47  |
| operator complexity | 1.9   |

aprokop:
Branch to fix #31.

### DIFF
--- a/tests/test_hierarchy.cc
+++ b/tests/test_hierarchy.cc
@@ -209,6 +209,6 @@ BOOST_AUTO_TEST_CASE(hierarchy_2d)
     res[i + 1] = rel_residual;
   }
 
-  pcout << "Convergence rate: " << res[n_cycles] / res[n_cycles - 1]
-        << std::endl;
+  pcout << "Convergence rate: " << std::fixed << std::setprecision(2)
+        << res[n_cycles] / res[n_cycles - 1] << std::endl;
 }


tests/test_hierarchy.cc
```diff
@@ -209,6 +209,6 @@ BOOST_AUTO_TEST_CASE(hierarchy_2d)
     res[i + 1] = rel_residual;
   }
 
-  pcout << "Convergence rate: " << res[n_cycles] / res[n_cycles - 1]
-        << std::endl;
+  pcout << "Convergence rate: " << std::fixed << std::setprecision(2)
```
Rombur: You should removed the `std::scientific` from line 198 if you want `std::fixed`

aprokop: I want scientific for residuals, but fixed for the rate.

PR #34, opened 2018-03-08
Fix a bug when using multiple eigenvectors per agglomerate

Rombur:
With two eigenvalues we get
- Grid complexity: 1.47
- Operator complexity:  1.9
- Convergence rate: 0.1

So we have the same results as them except for the convergence rate. We have 0.10 and they have 0.04. 

Relates to #31


Rombur:
Yes it is on my todo list

PR #38, opened 2018-03-14
Add support for more geometries

Rombur:
This add  support  for lapack so that we can switch betwen lapack and arpack to compute the eigenvalues. The advantage of lapack is that we don't require to invert the local system which means there is no problem when the local system is singular (e.g. Laplace without boundary condition). Also add support for hyper_ball and meshes which vertices are randomly moved.

Still missing a test case.

Example of new mesh (interior cells are distorted in purpose)
distorted_disc

### DIFF
--- a/include/mfmg/adapters_dealii.hpp
+++ b/include/mfmg/adapters_dealii.hpp
@@ -156,7 +156,7 @@ public:
                    std::shared_ptr<boost::property_tree::ptree> params)
   {
     AMGe_host<mesh_type::dimension(), mesh_evaluator_type, vector_type> amge(
-        comm, mesh._dof_handler, params->get("eigensolver: arpack", true));
+        comm, mesh._dof_handler, params->get("eigensolver: type", "arpack"));
 
     std::array<unsigned int, dim> agglomerate_dim;
     agglomerate_dim[0] = params->get<unsigned int>("agglomeration: nx");
--- a/include/mfmg/adapters_dealii.hpp
+++ b/include/mfmg/adapters_dealii.hpp
@@ -156,7 +156,7 @@ public:
                    std::shared_ptr<boost::property_tree::ptree> params)
   {
     AMGe_host<mesh_type::dimension(), mesh_evaluator_type, vector_type> amge(
-        comm, mesh._dof_handler);
+        comm, mesh._dof_handler, params->get("eigensolver: arpack", true));
 
     std::array<unsigned int, dim> agglomerate_dim;
     agglomerate_dim[0] = params->get<unsigned int>("agglomeration: nx");
--- a/tests/laplace.hpp
+++ b/tests/laplace.hpp
@@ -40,7 +40,8 @@ public:
 
   void setup_system(int num_refinements = 3);
 
-  void assemble_system(dealii::Function<dim> const &source);
+  void assemble_system(dealii::Function<dim> const &source,
+                       dealii::Function<dim> const &material_property);
 
   template <typename PreconditionerType>
   VectorType solve(PreconditionerType &preconditioner);
@@ -51,7 +52,8 @@ public:
 
   template <typename PreconditionerType>
   void run(PreconditionerType &preconditioner,
-           dealii::Function<dim> const &source);
+           dealii::Function<dim> const &source,
+           dealii::Function<dim> const &material_property);
 
   // The following variable should be private but there are public for
   // simplicity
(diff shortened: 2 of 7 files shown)


include/mfmg/adapters_dealii.hpp
```diff
@@ -156,7 +156,7 @@ class Adapter<DealIIMeshEvaluator<dim, VectorType>>
                    std::shared_ptr<boost::property_tree::ptree> params)
   {
     AMGe_host<mesh_type::dimension(), mesh_evaluator_type, vector_type> amge(
-        comm, mesh._dof_handler);
+        comm, mesh._dof_handler, params->get("eigensolver: arpack", true));
```
aprokop: Would prefer a string parameter "eigensolver: type" or similar, with possible values "arpack" or "lapack". So that in the future, we won't have to change it when we introduce our own solver.

tests/laplace.hpp
```diff
@@ -75,13 +84,36 @@ Laplace<dim, VectorType>::Laplace(MPI_Comm const &comm, unsigned int fe_degree)
...
+    dealii::GridGenerator::hyper_ball(_triangulation);
+    _triangulation.set_all_manifold_ids_on_boundary(0);
+    _triangulation.set_manifold(0, _spherical_manifold);
+  }
+  else
+    dealii::GridGenerator::hyper_cube(_triangulation);
+
+  _triangulation.refine_global(ptree.get("n_refinements", 3));
+
+  if (ptree.get("distort_random", false))
```
aprokop: Can you please add "distort_random" to the input deck?

tests/laplace.hpp
```diff
@@ -75,13 +84,36 @@ Laplace<dim, VectorType>::Laplace(MPI_Comm const &comm, unsigned int fe_degree)
...
+    dealii::GridTools::distort_random(0.2, _triangulation);
 
   _dof_handler.distribute_dofs(_fe);
 
+  std::string const renumbering = ptree.get("renumbering", "None");
```
aprokop: Please change "renumbering" to more commonly used "reordering" and add it to the input deck.

tests/laplace.hpp
```diff
@@ -38,9 +43,10 @@ class Laplace
 public:
   Laplace(MPI_Comm const &comm, unsigned int fe_degree);
 
-  void setup_system(int num_refinements = 3);
+  void setup_system(boost::property_tree::ptree const ptree);
```
aprokop: Would it not be better to add parameters to the constructor arguments, store them and then use for `setup_system` and `assemble_system`?

tests/test_hierarchy.cc
```diff
@@ -47,6 +47,107 @@ double Source<dim>::value(dealii::Point<dim> const &, unsigned int const) const
...
+
+  virtual double value(dealii::Point<dim> const &p,
+                       unsigned int const component = 0) const override final;
+};
+
+template <int dim>
+double ConstantMaterialProperty<dim>::value(dealii::Point<dim> const &,
+                                            unsigned int const) const
+{
+  return 1.;
```
aprokop: If we are not using ETI inside the tests, is there a reason to move the body of the member functions out of declaration?

tests/test_hierarchy.cc
```diff
@@ -47,6 +47,107 @@ double Source<dim>::value(dealii::Point<dim> const &, unsigned int const) const
...
+                       unsigned int const component = 0) const override final;
+};
+
+template <int dim>
+double LinearMaterialProperty<dim>::value(dealii::Point<dim> const &p,
+                                          unsigned int const) const
+{
+  double value = 1.;
+  for (unsigned int d = 0; d < dim; ++d)
+    value += p[d];
```
aprokop: Maybe add some changing coefficients for each dimension, so we don't get lucky accidentally.

tests/test_hierarchy.cc
```diff
@@ -47,6 +47,107 @@ double Source<dim>::value(dealii::Point<dim> const &, unsigned int const) const
...
+};
+
+template <int dim>
+double DiscontinuousMaterialProperty<dim>::value(dealii::Point<dim> const &p,
+                                                 unsigned int const) const
+{
+  double value = 10.;
+  for (unsigned int d = 0; d < dim; ++d)
+    if (p[d] > 0.5)
+      value *= value;
```
aprokop: A more challenging problem would be chess coloring. I also remember some multigrid methods (blackbox?) were tested on some problem where they made the discontinuity not grid aligned. Will need to find it out (I think it may be in the Proceedings of DD20).

tests/laplace.hpp
Rombur: Done

tests/laplace.hpp
Rombur: I have no opinion on this. `ptree` is only used in `setup_system` and `MaterialProperty` in `assemble_system`. I can add them in the constructor if you prefer.

PR #41, opened 2018-03-16
Fix a bug in a test (test was not using property tree)

Rombur:
I don't know how this is possible but master is broken. This should fix the problem.


Rombur:
Since this is a bug for master, I will merge myself.

Rombur:
Alright master is passing again. Catastrophe avoided (users were not impacted :laughing: )

aprokop:
I guess the only way master could be broken if we have two merge commits with the second one branched out before first one merged.

Issue #37, opened 2018-03-09
Clean up


Rombur:
Here is a list of things that we can do to improve the code:
  - [ ] Replace `class` by `typename` in template parameter (right now we are not using any thing consistently. Example: `VectorType is sometimes a `class` sometimes a `typename`)
  - [ ] Put `const` on the right
  - [ ] Rename `dealii_operator` to `operator_deali`i (consistency with `adapter_dealii`)
  - [ ] Use child property tree
  - [ ] More tests: including a test with variable number of eigenvectors per agglomerate.
  - [ ] Reorder `public`, `protected`, and `private`
  - [ ] `evaluate` for the `GlobalOperator` requires to copy the system sparse matrix. This is a pretty expensive operation. Instead, we should just ask for a shared pointer

PR #42, opened 2018-03-16
Operator refactor

Rombur:
The first two commits relates to #37 and are mostly cosmetic. The third commit is more interesting. I have talked with @dalg24 and we thought that the current design was too complicated. This PR removes `DealIIOperator` (which was missing some functions and therefore was not an `Operator` as defined in `concept.hpp`) and it also removes `DealIIMatrixOperator`. Instead of specializing `DealIIMatrixOperator`, we now have classes that derived directly from `Operator`. The classes are `DealIIMatrixOperator` for the `dealii::SparseMatrix` and `TrilinosMatrixOperator` for the `dealii::TrilinosWrappers::SparseMatrix`. This classes are in the `dealii_adapter` namespace and should probably be moved in a `dealii` subdirectory later. The new structure is simpler and allows us to use ETI.

### DIFF
--- a/include/mfmg/concepts.hpp
+++ b/include/mfmg/concepts.hpp
@@ -14,12 +14,15 @@
 
 #include <mfmg/exceptions.hpp>
 
+#include <boost/property_tree/ptree.hpp>
+
 #include <memory>
+#include <mpi.h>
 
 namespace mfmg
 {
 
-template <class VectorType>
+template <typename VectorType>
 class Operator
 {
 public:
@@ -27,16 +30,19 @@ public:
   using vector_type = VectorType;
 
 public:
-  void apply(const vector_type &x, vector_type &y) const;
-  std::shared_ptr<operator_type> transpose() const;
-  std::shared_ptr<operator_type> multiply(const operator_type &b) const;
+  virtual size_t m() const = 0;
+  virtual size_t n() const = 0;
+  virtual void apply(vector_type const &x, vector_type &y) const = 0;
+  virtual std::shared_ptr<operator_type> transpose() const = 0;
+  virtual std::shared_ptr<operator_type>
+  multiply(operator_type const &b) const = 0;
 
-  std::shared_ptr<vector_type> build_domain_vector() const;
-  std::shared_ptr<vector_type> build_range_vector() const;
+  virtual std::shared_ptr<vector_type> build_domain_vector() const = 0;
+  virtual std::shared_ptr<vector_type> build_range_vector() const = 0;
 };
 
-template <class Mesh, class GlobalOperatorType,
-          class LocalOperatorType = GlobalOperatorType>
+template <typename Mesh, typename GlobalOperatorType,
+          typename LocalOperatorType = GlobalOperatorType>
 class MeshEvaluator
 {
 public:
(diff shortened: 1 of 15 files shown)


include/mfmg/concepts.hpp
```diff
@@ -14,29 +14,35 @@
...
   using vector_type = VectorType;
 
 public:
-  void apply(const vector_type &x, vector_type &y) const;
-  std::shared_ptr<operator_type> transpose() const;
-  std::shared_ptr<operator_type> multiply(const operator_type &b) const;
+  virtual size_t m() const = 0;
+  virtual size_t n() const = 0;
+  virtual void apply(vector_type const &x, vector_type &y) const = 0;
+  virtual std::shared_ptr<operator_type> transpose() const = 0;
```
aprokop: The concept of operator should really only have the meaning of how to produce right hand side by operating on left hand side. Its interface should clearly indicate that. Thus, I believe that member functions like `transpose()` and `multiply()` do not belong in it, as they somewhat restrict the concept to operators corresponding to matrices. It also becomes a rabbit hole if we start creating operators with more required functions and putting them in `Operator`. Therefore, I'd like to retain the old interface with `apply` and `build_{range,domain}_vector`. We could even take `m()` and `n()` out but that would most likely require `Vector` concept.

aprokop: What I would propose to do instead is to create a concept of `MatrixOperator` which inherits `Operator`. Then, in `DealIIMatrixOperator` and `TrilinosMatrixOperator` could inherit from that instead.

include/mfmg/dealii_operator.hpp
```diff
@@ -28,413 +31,161 @@
...
-  }
-};
-
-template <class VectorType>
-class DealIIMatrixOperator<
-    dealii::SparsityPattern,
-    dealii::SparseMatrix<typename VectorType::value_type>, VectorType>
-    : public DealIIOperator<VectorType>
+template <typename VectorType>
+class DealIIMatrixOperator : public Operator<VectorType>
```
aprokop: As I mentioned above, this should inherit from `MatrixOperator` concept.

include/mfmg/dealii_operator.hpp
```diff
@@ -28,413 +31,161 @@
...
+
   std::shared_ptr<matrix_type> _matrix;
 };
 
-template <class VectorType>
-class DealIIMatrixOperator<dealii::TrilinosWrappers::SparsityPattern,
-                           dealii::TrilinosWrappers::SparseMatrix, VectorType>
-    : public DealIIOperator<VectorType>
+template <typename VectorType>
+class TrilinosMatrixOperator : public Operator<VectorType>
```
aprokop: I propose either renaming `TrilinosMatrixOperator` to `DealIITrilinosMatrixOperator` or dropping `DealII` from `DealIIMatrixOperator` for consistency. They both live in some `dealii` namespace so that should be fine. Judging of how this patch renames `DealIISmootherOperator` to `SmootherOperator`, I am fine changing `DealIIMatrixOperator` to simply `MatrixOperator`.

include/mfmg/dealii_operator.hpp
```diff
@@ -28,413 +31,161 @@
...
-        _jacobi_smoother.vmult(x, b);
-        break;
-      case ILU:
-        _ilu_smoother.vmult(x, b);
-        break;
-      };
-    }
-  }
+  virtual std::shared_ptr<operator_type>
+  multiply(operator_type const &operator_b) const override final;
```
aprokop: These functions really don't belong to smoother, and should be gone once `MatrixOperator` is moved out of `Operator` concept.

include/mfmg/dealii_operator.templates.hpp
```diff
@@ -0,0 +1,296 @@
...
+  ASSERT_THROW_NOT_IMPLEMENTED();
+
+  return nullptr;
+}
+
+template <typename VectorType>
+std::shared_ptr<VectorType>
+DealIIMatrixOperator<VectorType>::build_domain_vector() const
+{
+  ASSERT_THROW_NOT_IMPLEMENTED();
```
aprokop: `build_domain_vector` and `build_range_vector` should probably be implemented for `DealIIMatrixOperator`. I guess they currently are not only because we don't use them.

include/mfmg/dealii_operator.hpp
```diff
@@ -28,413 +31,161 @@
...
-  const matrix_type &_matrix;
-  SmootherType _prec_type;
-  dealii::TrilinosWrappers::PreconditionSSOR _sgs_smoother;
-  dealii::TrilinosWrappers::PreconditionSOR _gs_smoother;
-  dealii::TrilinosWrappers::PreconditionJacobi _jacobi_smoother;
-  dealii::TrilinosWrappers::PreconditionILU _ilu_smoother;
+  void initialize(std::string const &prec_type);
+
+  matrix_type const &_matrix;
+  std::unique_ptr<dealii::TrilinosWrappers::PreconditionBase> _smoother;
```
aprokop: OK, that was be being stupid, I did not find Trilinos' `PreconditionerBase` in deal.ii.

include/mfmg/dealii_operator.hpp
```diff
@@ -28,413 +31,161 @@
...
 private:
-  const matrix_type &_matrix;
-  SmootherType _prec_type;
-  dealii::TrilinosWrappers::PreconditionSSOR _sgs_smoother;
-  dealii::TrilinosWrappers::PreconditionSOR _gs_smoother;
-  dealii::TrilinosWrappers::PreconditionJacobi _jacobi_smoother;
-  dealii::TrilinosWrappers::PreconditionILU _ilu_smoother;
+  void initialize(std::string const &prec_type);
+
+  matrix_type const &_matrix;
```
aprokop: We need to add a comment here. Something like this:
```
Currently, dealii preconditioners do not support a mode to update a given vector, instead replacing it. For smoothers, this mode is necessary to operator. We workaround this issue by doing extra compuations in `apply`, however this requires residual computation. Therefore, we need to store a reference to matrix.
```

include/mfmg/concepts.hpp
Rombur: Yeah, good idea.

include/mfmg/dealii_operator.hpp
Rombur: But then how do you call the matrix operator that derives from `Operator`? I have used `TrilinosMatrixOperator` and `DealIIMatrixOperator` because when Damien saw `DealIIMatrixOperator`, he assumed that we had sth called `TrilinosMatrixOperator`

include/mfmg/dealii_operator.templates.hpp
Rombur: Yes, I think it's fine for now not having them

include/mfmg/dealii_operator.hpp
aprokop: Yeah, while we could differentiate by namespaces, I like calling the concept of matrix operator `MatrixOperator`. The only minor issue I have right now is that `TrilinosMatrixOperator` does not indicate that it is deal.ii specific. I guess the namespace does that, but that's a bit of extra effort. I could live with leaving it as it is, but would prefer `DealIITrilinosMatrixOperator` and then probably `DealIIDirectOperator` with `DealIISmootherOperator`. In any case, this all is hidden from Hierarchy.

PR #43, opened 2018-03-18
Improve ReadTheDocs documentation

Rombur:
Not a very interesting patch but we can't put on the report that we use `ReadTheDocs` if there is no useful information on the website.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -1,0 +1,57 @@
+Installation
+============
+
+This section provide guidelines for installing DataTransferKit and its TPLs.
+
+Install third-party libraries
+-----------------------------
+
+The following third party libraries (TPLs) are used by mfmg:
+
++------------------------+---------+
+| Packages               | Version |
++========================+=========+
+| ARPACK                 | N/A     |
++------------------------+---------+
+| Boost                  | 1.65.1  |
++------------------------+---------+
+| BLAS/LAPACK            | N/A     |
++------------------------+---------+
+| deal.II                | 8.5     |
++------------------------+---------+
+| MPI                    | N/A     |
++------------------------+---------+
+| Trilinos               | 12.X    |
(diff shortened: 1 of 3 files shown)


doc/install.rst
```diff
@@ -0,0 +1,57 @@
...
++------------------------+---------+
+| Packages               | Version |
++========================+=========+
+| ARPACK                 | N/A     |
++------------------------+---------+
+| Boost                  | 1.65.1  |
++------------------------+---------+
+| BLAS/LAPACK            | N/A     |
++------------------------+---------+
+| deal.II                | 8.5     |
```
aprokop: Is 8.5 sufficient? I thought we need `develop`?

doc/install.rst
```diff
@@ -0,0 +1,57 @@
...
++------------------------+---------+
+| MPI                    | N/A     |
++------------------------+---------+
+| Trilinos               | 12.X    |
```
aprokop: If deal.ii uses only Epetra, I assume that it works with a bunch earlier versions too?

doc/install.rst
```diff
@@ -0,0 +1,57 @@
...
++------------------------+---------+
+
+The dependencies of mfmg may be built using `Spack
+<https://github.com/llnl/spack>`_ package manager. You need to install the
+following package:
+
+.. code::
+
+    $ spack install dealii
```
aprokop: I remember you mentioning `dealii@develop`? That's what I've been installing. Or is it only to get some GPU in?

doc/install.rst
Rombur: Yeah it should with older versions but I don't remember which one. It won't work with 11.2 for example.

doc/install.rst
Rombur: It's only for GPU

PR #46, opened 2018-04-05
Reduce tolerance when testing the hierarchy on hyper_ball

### LINKED ISSUE #40: Convergence on hyperball
Convergence on hyperball is degraded on the testing machine. We need to investigate why. A good first step would be to reproduce the results with Docker

Rombur:
Close #40 The bad convergence disappears when updating the Docker image to the latest deal.II version. There was some work in deal.II to improve the `hyper_ball` mesh, so this is probably what I saw.


Rombur:
Yes the image was pushed on dockerhub. If it wasn't, jenkins would not have passed because we always try to pull the latest image from dockerhub

PR #47, opened 2018-04-05
Update Docker image to use a newer version of deal.II

Rombur:
Relates to #46

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -187,7 +187,7 @@ RUN export ARPACK_VERSION=3.5.0 && \
 ENV ARPACK_DIR=${INSTALL_DIR}/arpack
 
 # Install deal.II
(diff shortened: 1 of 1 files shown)


ci/Dockerfile
```diff
@@ -205,6 +205,8 @@ RUN export DEAL_II_HASH=7ac7850030de7a2027679099073f00f59a076ac4 && \
         -DTRILINOS_DIR=${TRILINOS_DIR} \
         -DDEAL_II_WITH_ARPACK=ON \
         -DARPACK_DIR=${ARPACK_DIR} \
+        -DDEAL_II_WITH_CUDA=ON \
+        -DDEAL_II_WITH_CXX14=OFF \
```
aprokop: Is turning C++14 off required for Cuda?

Rombur: Yes exactly. deal.II does not support compiling the CUDA code with a different C++ standard than the rest of the library.

PR #54, opened 2018-04-18
Add SparseMatrixDeviceOperator

Rombur:
This adds the `SparseMatrixDevice` class. It's a pretty simple wrappers around `SparseMatrixDevice`. The main difficulty is to create the transpose of the matrix. Since this is an operation that's done only once, I move all the data to the host  then use `Epetra` to transpose the matrix and finally, I move it back to the device.

The first commit also fix a bug where the wrong amount of memory was allocated when moving an `Epetra_CrsMatrix` to the device.

Something I just realized is that I have implemented everything `dealii_operator_device.cuh` in analogy to `dealii_operator_device` but these classes don't really depend on `deal.II` so it might not be the best name.

PS: I'd like to have this PR merged today because the tester will be down for five days starting tomorrow.

### DIFF
--- a/include/mfmg/dealii_operator_device.templates.cuh
+++ b/include/mfmg/dealii_operator_device.templates.cuh
@@ -23,6 +23,127 @@
 namespace mfmg
 {
 template <typename VectorType>
+SparseMatrixDeviceOperator<VectorType>::SparseMatrixDeviceOperator(
+    std::shared_ptr<SparseMatrixDevice<typename VectorType::value_type>> matrix)
+    : _matrix(matrix)
+{
+  ASSERT(matrix != nullptr, "The matrix must exist");
+}
+
+template <typename VectorType>
+void SparseMatrixDeviceOperator<VectorType>::apply(VectorType const &x,
+                                                   VectorType &y) const
+{
+  _matrix->vmult(y, x);
+}
+
+template <typename VectorType>
+std::shared_ptr<MatrixOperator<VectorType>>
+SparseMatrixDeviceOperator<VectorType>::transpose() const
+{
+  // TODO: to do this on the GPU
+  // Copy the data to the cpu and then, use the trilinos to compute the
+  // transpose. This is not the most efficient way to do this but it is the
+  // easiest.
+  unsigned int const local_nnz = _matrix->local_nnz();
+  unsigned int const n_local_rows = _matrix->n_local_rows();
+  std::vector<value_type> values(local_nnz);
+  std::vector<int> column_index(local_nnz);
+  std::vector<int> row_ptr(n_local_rows + 1);
+
+  // Copy the data to the host
+  cuda_mem_copy_to_host(_matrix->val_dev, values);
+  cuda_mem_copy_to_host(_matrix->column_index_dev, column_index);
+  cuda_mem_copy_to_host(_matrix->row_ptr_dev, row_ptr);
+
+  // Create the sparse matrix on the host
+  dealii::IndexSet locally_owned_rows = _matrix->locally_owned_range_indices();
+  dealii::TrilinosWrappers::SparseMatrix sparse_matrix(
+      locally_owned_rows, _matrix->locally_owned_domain_indices(),
+      _matrix->get_mpi_communicator());
+
+  std::vector<unsigned int> rows;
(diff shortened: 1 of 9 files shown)


include/mfmg/dealii_operator_device.templates.cuh
```diff
@@ -22,6 +22,127 @@
...
+      _matrix->get_mpi_communicator());
+
+  std::vector<unsigned int> rows;
+  locally_owned_rows.fill_index_vector(rows);
+  for (unsigned int i = 0; i < n_local_rows; ++i)
+  {
+    unsigned int const row = rows[i];
+    unsigned int const n_cols = row_ptr[i + 1] - row_ptr[i];
+    for (unsigned int j = row_ptr[i]; j < row_ptr[i + 1]; ++j)
+      sparse_matrix.set(row, column_index[j], values[j]);
```
aprokop: Is there no way in dealii to construct a matrix from 3 pointers? Or at least row by row instead of entry by entry? This seems tedious.

include/mfmg/sparse_matrix_device.templates.cuh
```diff
@@ -266,6 +303,7 @@ void SparseMatrixDevice<ScalarType>::mmult(
   // TODO Communicate the values
 
   // Compute the number of non-zero elements in C
+  ASSERT(B.m() == n(), "The matrices cannot be multiplied together");
```
aprokop: Improve the error message by also printing the sizes of matrices you are trying to multiply.

tests/test_sparse_matrix_device_operator.cu
```diff
@@ -0,0 +1,145 @@
...
+BOOST_AUTO_TEST_CASE(matrix_operator)
+{
+  // Create the cusparse_handle
+  cusparseHandle_t cusparse_handle = nullptr;
+  cusparseStatus_t cusparse_error_code;
+  cusparse_error_code = cusparseCreate(&cusparse_handle);
+  mfmg::ASSERT_CUSPARSE(cusparse_error_code);
+
+  unsigned int const n_rows = 30;
+  unsigned int const nnz_per_row = 1;
```
aprokop: Probably want to change it to something greater than 1. Otherwise, `n_rows` = `n_cols`, and the checks for `domain` and `range` can hide a problem.

tests/test_sparse_matrix_device_operator.cu
```diff
@@ -0,0 +1,145 @@
...
+  dealii::Vector<double> transpose_domain_dealii(
+      transpose_domain_vector.begin(), transpose_domain_vector.end());
+  dealii::Vector<double> transpose_range_dealii(transpose_matrix_operator->m());
+  transpose_sparse_matrix.vmult(transpose_range_dealii,
+                                transpose_domain_dealii);
+  std::vector<double> transpose_range_vector(transpose_matrix_operator->m());
+  mfmg::cuda_mem_copy_to_host(transpose_range_dev->val_dev,
+                              transpose_range_vector);
+  for (unsigned int i = 0; i < n_rows; ++i)
+    BOOST_CHECK_EQUAL(transpose_range_vector[i], transpose_range_dealii[i]);
```
aprokop: From what I see, it seems that you check a transposed matrix by checking the result of `apply`. However, what if `build_domain_vector()` initializes it to 0? Then you would get 0 in both cases but the matrix may not be correct. The approach is fine, but probably need to add an explicit randomization setup.

tests/test_sparse_matrix_device_operator.cu
```diff
@@ -0,0 +1,145 @@
...
+
+  // Check multiply
+  sparse_matrix.vmult(range_dealii, transpose_range_dealii);
```
aprokop: Similar to above, I would like to have explicit calls to random here to make sure.

tests/test_sparse_matrix_device_operator.cu
Rombur: Yeah I had a bug that I was trying to fix and I forgot to change the value back.

tests/test_sparse_matrix_device_operator.cu
Rombur: The vector is initialized to 1.

Rombur:
OK done

PR #55, opened 2018-04-20
Add SmootherDeviceOperator

Rombur:
This is the last missing piece to implement all the operators. First three commits are for helper functions. Note that the implementation is not optimized (I use a matrix to store the inverse diagonal) but it is good enough for now.

### DIFF
--- a/include/mfmg/dealii_operator_device.templates.cuh
+++ b/include/mfmg/dealii_operator_device.templates.cuh
@@ -20,6 +20,8 @@
 
 #include <algorithm>
 
+#define BLOCK_SIZE 512
+
 namespace mfmg
 {
 template <typename VectorType>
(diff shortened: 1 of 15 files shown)


Rombur:
_ping_

Rombur:
merge_pr

include/mfmg/dealii_operator_device.templates.cuh
```diff
@@ -145,6 +147,117 @@ SparseMatrixDeviceOperator<VectorType>::build_range_vector() const
...
+}
+
+template <typename VectorType>
+void SmootherDeviceOperator<VectorType>::apply(VectorType const &b,
+                                               VectorType &x) const
+{
+  // r = -(b - Ax)
+  vector_type r(b);
+  _matrix.vmult(r, x);
+  r.add(-1., b);
```
aprokop: Maybe implement an proper `alpha*A*x + beta*y` `apply` call for the device matrix? Then, this can be done more efficiently.

include/mfmg/dealii_operator_device.templates.cuh
```diff
@@ -145,6 +147,117 @@ SparseMatrixDeviceOperator<VectorType>::build_range_vector() const
...
+template <typename VectorType>
+void SmootherDeviceOperator<VectorType>::initialize(std::string &prec_type)
+{
+  // Transform to lower case
+  std::transform(prec_type.begin(), prec_type.end(), prec_type.begin(),
+                 tolower);
+
+  ASSERT_THROW(prec_type == "jacobi", "Only Jacobi smoother is implemented.");
+
+  ASSERT(_matrix.m() == _matrix.n(), "The matrix is not square");
```
aprokop: print the actual dimensions of the matrix

include/mfmg/dealii_operator_device.templates.cuh
```diff
@@ -145,6 +147,117 @@ SparseMatrixDeviceOperator<VectorType>::build_range_vector() const
...
+  unsigned int const size = _matrix.n_local_rows();
+  value_type *val_dev = nullptr;
+  cuda_malloc(val_dev, size);
+  int *column_index_dev = nullptr;
+  cuda_malloc(column_index_dev, size);
+  int *row_ptr_dev = nullptr;
+  cuda_malloc(row_ptr_dev, size + 1);
+  unsigned int const local_nnz = _matrix.local_nnz();
+  int *row_index_dev = nullptr;
+  cuda_malloc(row_index_dev, local_nnz);
```
aprokop: I don't understand what's going on here. There is a mix of sizes. Maybe reorganize, or which variables correspond to what? From the names, it's not clear what is `row_ptr_dev` and `row_index_dev`, or why the later is `local_nnz` size. Maybe, use `triple` or `coo` in var names?

include/mfmg/dealii_operator_device.templates.cuh
```diff
@@ -145,6 +147,117 @@ SparseMatrixDeviceOperator<VectorType>::build_range_vector() const
...
+  ASSERT_CUSPARSE(cusparse_error_code);
+
+  int n_blocks = 1 + (local_nnz - 1) / BLOCK_SIZE;
+  internal::extract_inv_diag<<<n_blocks, BLOCK_SIZE>>>(
+      _matrix.val_dev, _matrix.column_index_dev, row_index_dev, local_nnz,
+      val_dev);
+
+  iota<<<n_blocks, BLOCK_SIZE>>>(size, column_index_dev);
+
+  n_blocks = 1 + size / BLOCK_SIZE;
```
aprokop: If `size` is `BLOCK_SIZE`, should `n_blocks` be 1 or 2?

include/mfmg/sparse_matrix_device.templates.cuh
```diff
@@ -241,6 +241,37 @@ SparseMatrixDevice<ScalarType>::SparseMatrixDevice(
...
 
+template <typename ScalarType>
+void SparseMatrixDevice<ScalarType>::reinit(
+    MPI_Comm comm, ScalarType *val_dev_, int *column_index_dev_,
+    int *row_ptr_dev_, unsigned int local_nnz,
+    dealii::IndexSet const &range_indexset,
+    dealii::IndexSet const &domain_indexset, cusparseHandle_t cusparse_handle_)
+{
+  // This function can only be called if the object is empty. Otherwise we need
+  // to deal with the ownership of the data
```
aprokop: This seems like it will cause problems at some points. Can we assert this?

include/mfmg/vector_device.templates.cuh
```diff
@@ -0,0 +1,73 @@
...
+ *************************************************************************/
+
+#ifndef VECTOR_DEVICE_TEMPLATES_CUH
+#define VECTOR_DEVICE_TEMPLATES_CUH
+
+#ifdef MFMG_WITH_CUDA
+
+#include <mfmg/vector_device.cuh>
+
+#define BLOCK_SIZE 512
```
aprokop: Should `BLOCK_SIZE` be a configure-time global MFMG constant?

include/mfmg/vector_device.templates.cuh
```diff
@@ -0,0 +1,73 @@
...
+
+namespace mfmg
+{
+namespace internal
+{
+template <typename ScalarType>
+__global__ void add(int const size, ScalarType alpha,
```
aprokop: Maybe generalize to `alpha*x_val + beta*val`?

include/mfmg/dealii_operator_device.templates.cuh
Rombur: Let me open an issue for this because I plan to switch later to the deal.II distributed vectors. So I don't want to implement something that we won't use for a long time.

include/mfmg/dealii_operator_device.templates.cuh
Rombur: It doesn't really matter. We will launch a bunch of threads that won't do any work.

include/mfmg/vector_device.templates.cuh
Rombur: We could but I don't think it really matters. Basically you need to pick 256 or 512. This number comes from the way Nvidia designs its gpus.

aprokop: OK, so it does not have to be a configure-time. But, I saw you declare it in multiple places. Could you put it in one place, and use throughout?

Rombur: Yes I will change it.

include/mfmg/sparse_matrix_device.templates.cuh
Rombur: done

Rombur:
I am not sure what's going on with the tester. Everything works fine locally. I am investigating.

Rombur:
I have the same problem with master on fetnat. Something has changed since the reboot...

Rombur:
I found the problem but I don't know how to fix it. The tests that fail are using CUDA and MPI. In these tests, we are using both GPUs on fetnat and there is a problem with `device 1`. `malloc` segfaults on `device 1` even in serial. All the other tests pass because we are always using `device 0`.

ping @dalg24

Rombur:
Looks like a problem with the card
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 390.30                 Driver Version: 390.30                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla K40m          Off  | 00000000:04:00.0 Off |                    0 |
| N/A   22C    P8    19W / 235W |     11MiB / 11441MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Tesla K40m          Off  | 00000000:82:00.0 Off |                    3 |
| N/A   23C    P8    19W / 235W |     11MiB / 11441MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
```
and https://devtalk.nvidia.com/default/topic/993607/uncorrected-ecc-how-to-get-back-on-track/?offset=4

Rombur:
@aprokop this can be merged

PR #63, opened 2018-05-02
Add ML option for coarse solver

aprokop:
It's running ML now, I can see the output. Next steps:
- [x] propagate parameters from `boost::property_tree` to `Teuchos::ParameterList`
  Control the coarse levels from input deck. Hopefully not too hard, as ML parameters are flat.
- [x] Fix operator and grid complexities
  If one uses AMG for coarse levels too, it needs to combine its complexity with amge
- [x] Double check whether we solve the right thing
  ~Should ML be used as a preconditioner, or as a standalone solver? Do we pass residual or rhs?~ ML can be use as either a preconditioner, or as a standalone with a single iteration. There is no need to compute residual, it can be thought of having a zero initial guess.
- [ ] Study the convergence


aprokop:
It's running ML now, I can see the output. Next steps:
- [ ] propagate parameters from `boost::property_tree` to `Teuchos::ParameterList`
  Control the coarse levels from input deck. Hopefully not too hard, as ML parameters are flat.
- [ ] Fix operator and grid complexities
  If one uses AMG for coarse levels too, it needs to combine its complexity with amge
- [ ] Double check whether we solve the right thing
  Should ML be used as a preconditioner, or as a standalone solver? Do we pass residual or rhs?
- [ ] Study the convergence

dalg24:
Try something like that
```C++
boost::property_tree::ptree ptree; // assume it contains the data structure
std::stringstream ss;
boost::property_tree::xml_parser::write_xml( ss, ptree );
auto parameter_list = 
// NOTE I couldn't find a static function to parse the stream in Teuchos :/
Teuchos::ParameterList dummy;
auto plist = dummy.getParametersFromXmlString( ss.str() );
```

aprokop:
I think this is ready for review. It runs and converges, but I have not designed the test yet as I need to study the convergence.

source/utils.cc
```diff
@@ -41,4 +42,79 @@ void matrix_market_output_file(
...
+
+static bool is_integer(const std::string &s)
+{
+  std::istringstream is(s);
+  int i;
+  is >> i;
+  return !is.fail() && is.eof();
+}
+
+static bool is_double(const std::string &s)
```
dalg24: I think `(boost::property_tree::ptree::get_value_optional<T>()` does what you wanted.

tests/test_utils.cc
```diff
@@ -0,0 +1,61 @@
...
+#include <Teuchos_ParameterList.hpp>
+#include <boost/property_tree/info_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+BOOST_AUTO_TEST_CASE(plist2ptree)
+{
+  boost::property_tree::ptree ptree;
+  Teuchos::ParameterList plist;
+
+  std::string info_string = " \n\
```
dalg24: You might want to consider using a raw string literal here:
```C++
std::string info_string = R"(
eigensolver
{
  "number of eigenvectors" 2
  tolerance 1e-14
}
)"
```

Rombur:
Can you add a test that uses ML.

aprokop:
@Rombur Of course, but I need to check few things first.

aprokop:
@dalg24 Thank you for the comments, that worked!

aprokop:
@Rombur Added ML test. Ready for final review.

include/mfmg/dealii_operator.hpp
```diff
@@ -122,9 +126,14 @@ class DealIISmootherOperator : public Operator<VectorType>
                          std::shared_ptr<boost::property_tree::ptree> params);
 
   virtual size_t m() const override final { return _matrix.m(); }
-
   virtual size_t n() const override final { return _matrix.n(); }
 
+  virtual size_t grid_complexity() const override final { return m(); }
+  virtual size_t operator_complexity() const override final
+  {
+    ASSERT_THROW_NOT_IMPLEMENTED();
```
Rombur: Need a return to avoid warning

aprokop: Damn, it was there! I guess I messed up when cleaning and rebasing.

PR #66, opened 2018-05-05
Add docker-compose file to ease launching

aprokop:


doc/developer_tools.rst
```diff
@@ -0,0 +1,43 @@
...
+
+.. code:: bash
+
+    [host]$ docker-compose exec mfmg_dev bash
+
+Configure, build, and test as you would usually do:
+
+.. code:: bash
+
+    [container]$ ./ci/compile_and_run.sh .
```
Rombur: Have you tried that? How does that work with cuda? docker-compose does not support nvidia-docker plugin version 1. The reason it works with DTK is because @dalg24 created a fancy python script that sets a bunch of variable correctly

aprokop: I tried this on Mac. I was able to launch docker compose and run this script, which got to 20% at which point I get "Internal compiler error". So it may be that I have not run it far enough to break with cuda...

PR #68, opened 2018-05-08
Hierarchy driver

aprokop:
I would also like to add parameter reading from a command line to override the provided input deck parameters.


aprokop:
Interesting, I have no clue why this fails.

Rombur:
Just close and reopen the PR. The laplace tests are not great because there can be some race conditions when removing the output files.

aprokop:
Is is possible to `fsync` after writing data, and then `barrier` to avoid race conditions?

aprokop:
@Rombur OK. Updated this PR with the fix.

tests/CMakeLists.txt
```diff
@@ -1,12 +1,15 @@
 INCLUDE(${CMAKE_SOURCE_DIR}/cmake/SetupTests.cmake)
 
+INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR})
+
```
Rombur: Why do you do this?

tests/test_hierarchy.cc
```diff
@@ -18,6 +18,8 @@
 #include <mfmg/dealii_adapters.hpp>
 #include <mfmg/hierarchy.hpp>
 
+#include <test_hierarchy_helpers.hpp>
+
```
Rombur: Is it the reason why you want to include the `tests` directory in CMake? You can just use `#include "test_hierarchy_helpers.hpp"` instead. That's what we do with `laplace.hpp`

tests/test_hierarchy_driver.cc
```diff
@@ -0,0 +1,121 @@
...
+    res[i + 1] = rel_residual;
+  }
+
+  double const conv_rate = res[n_cycles] / res[n_cycles - 1];
+  pcout << "Convergence rate: " << std::fixed << std::setprecision(2)
+        << conv_rate << std::endl;
+}
+
+int main(int argc, char *argv[])
+{
```
Rombur: Is this file a test or not? If it is not then don't call the file `test_hierarchy_driver` and if it is, then you need to add a test. Same with `test_hierarchy_driver_helper`

aprokop:
@Rombur Addressed your comments.

PR #70, opened 2018-05-14
Add amgx 

### LINKED ISSUE #62: Implement AMGX for coarse levels

Rombur:
Most of this PR consists in changing the order of the data to what is required by amgx. Starting from CSR, we first need move the columns corresponding to the locally owned rows at the beginning of the data structures. Then the rows that don't contain any halo element needs to be move first. This permutation of the rows also requires us to swap the corresponding columns. The set up works as follows: 
 1. we communicate the sparsity pattern to all the processors
 2. we now have enough information to compute the communication data structures
 3. once the communication pattern is known, we now which elements are part of the halo and thus we can reorder the matrix
 4. because the matrix has been change we need to update the communication pattern

When solving the system, we need to reorder the right hand side and the solution since the matrix has been changed. For now, the parameters for amgx are read from a different file because I add trouble given the parameter as a string.

Closes #62


Rombur:
The tests are failing for a reason unrelated to the PR

Rombur:
_ping_

Rombur:
_ping_

aprokop:
@Rombur sorry, did not realize you pushed all the changes.

Rombur:
need-you-to-merge-that-pull-request

Rombur:
Thanks

PR #71, opened 2018-05-15
Parallel mmult

Rombur:
In this PR, I've removed what was left of the `float` instantiation,  move code around to add a function that creates a `dealii::TrilinosWrappers::SparseMatrix` from a `SparseMatrixDevice`, and then use deal.II to do `mmult`. I have also moved around the destructor of `SparseMatrixDevice` because it was at the wrong place in the file but I haven't changed the code at all.

### DIFF
--- a/include/mfmg/sparse_matrix_device.cuh
+++ b/include/mfmg/sparse_matrix_device.cuh
@@ -48,6 +48,9 @@ public:
 
   ~SparseMatrixDevice();
 
+  SparseMatrixDevice<ScalarType> &
+  operator=(SparseMatrixDevice<ScalarType> &&other);
+
   /**
    * Reinitialize the matrix. This can only be called if the object it empty.
    */
--- a/include/mfmg/sparse_matrix_device.templates.cuh
+++ b/include/mfmg/sparse_matrix_device.templates.cuh
@@ -240,6 +240,65 @@ SparseMatrixDevice<ScalarType>::SparseMatrixDevice(
 }
 
 template <typename ScalarType>
+SparseMatrixDevice<ScalarType>::~SparseMatrixDevice()
+{
+  if (val_dev != nullptr)
+  {
+    cuda_free(val_dev);
+    val_dev = nullptr;
+  }
+
+  if (column_index_dev != nullptr)
+  {
+    cuda_free(column_index_dev);
+    column_index_dev = nullptr;
+  }
+
+  if (row_ptr_dev != nullptr)
+  {
+    cuda_free(row_ptr_dev);
+    row_ptr_dev = nullptr;
+  }
+
+  if (descr != nullptr)
+  {
+    cusparseStatus_t cusparse_error_code;
+    cusparse_error_code = cusparseDestroyMatDescr(descr);
+    ASSERT_CUSPARSE(cusparse_error_code);
+    descr = nullptr;
+  }
+}
+
+template <typename ScalarType>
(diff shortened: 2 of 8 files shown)


include/mfmg/sparse_matrix_device.cuh
```diff
@@ -48,6 +48,9 @@ public:
 
   ~SparseMatrixDevice();
 
+  SparseMatrixDevice<ScalarType> &
+  operator=(SparseMatrixDevice<ScalarType> &&other);
```
dalg24: Why are you providing a move assignment operator here?  You might want to do that in a separate commit.

include/mfmg/utils.cuh
```diff
@@ -51,6 +51,9 @@ convert_matrix(dealii::TrilinosWrappers::SparseMatrix const &sparse_matrix);
 SparseMatrixDevice<double>
 convert_matrix(Epetra_CrsMatrix const &sparse_matrix);
 
+dealii::TrilinosWrappers::SparseMatrix
+convert_matrix(SparseMatrixDevice<double> const &matrix_dev);
```
dalg24: Same name for conversion `Epetra_CrsMatrix` -> `SparseMatrixDevice` and `SparseMatrixDevice` -> `dealii::TrilinosWrappers::SparseMatrix` is a confusing.

source/utils.cu
```diff
@@ -167,6 +167,42 @@ SparseMatrixDevice<double> convert_matrix(Epetra_CrsMatrix const &sparse_matrix)
...
+  for (auto row : locally_owned_rows)
+  {
+    unsigned int const n_cols = row_ptr[pos + 1] - row_ptr[pos];
+    sparse_matrix.set(
+        row, n_cols,
+        reinterpret_cast<unsigned int *>(column_index.data() + row_ptr[pos]),
+        values.data() + row_ptr[pos]);
+    ++pos;
+  }
+  sparse_matrix.compress(dealii::VectorOperation::insert);
```
dalg24: Would you please add a comment saying that ideally we would populate directly the three vectors (row pointer, column indices, and values)?

Rombur: You can't do that though. There is no way to do that with deal, this is the best you can do.

include/mfmg/sparse_matrix_device.cuh
Rombur: Because in `mmult` I need to do `C = convert_matrix()` and I don't want to copy the matrix.

include/mfmg/sparse_matrix_device.templates.cuh
```diff
@@ -341,38 +370,56 @@ void SparseMatrixDevice<ScalarType>::mmult(
...
+    ASSERT_CUSPARSE(cusparse_error_code);
+
+    // Reinitialize part of C
+    cuda_free(C.val_dev);
+    cuda_free(C.column_index_dev);
+    cuda_malloc(C.val_dev, C_local_nnz);
+    cuda_malloc(C.column_index_dev, C_local_nnz);
+    C._local_nnz = C_local_nnz;
+    C._nnz = C._local_nnz;
+    dealii::Utilities::MPI::sum(C._nnz, C._comm);
```
aprokop: You don't need it as `comm_size == 1`, right?

aprokop:
(@Rombur next time, could you just push the commits addressing the comments on top of the branch without rebasing? That would make it easier to see how comments were addressed).

Rombur:
Sure

PR #73, opened 2018-05-25
Material property

Rombur:
- Fix a bug where the local operators always used a diffusion coefficient of 1
- Make discontinuous problem harder to solve

### DIFF
--- a/tests/test_hierarchy_helpers.hpp
+++ b/tests/test_hierarchy_helpers.hpp
@@ -101,12 +101,11 @@ public:
                        unsigned int const = 0) const override final
 
   {
-    double value = 10.;
+    unsigned int dim_scale = 0;
     for (unsigned int d = 0; d < dim; ++d)
-      if (p[d] > 0.5)
-        value *= value;
+      dim_scale += static_cast<unsigned int>(std::floor(p[d] * 100)) % 2;
 
-    return value;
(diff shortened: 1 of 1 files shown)


tests/test_hierarchy_helpers.hpp
```diff
@@ -102,9 +102,19 @@ class DiscontinuousMaterialProperty : public dealii::Function<dim>
 
   {
     double value = 10.;
-    for (unsigned int d = 0; d < dim; ++d)
-      if (p[d] > 0.5)
-        value *= value;
+    for (unsigned int i = 0; i < 100; i += 2)
```
aprokop: Where is this hardcoded value of 100 coming from?

Rombur: I create a checkerboard of 100 by 100.

aprokop: Oh, I think I see what you are doing. Wouldn't it be easier if instead of looping, you would just do something like this:
```c++
unsigned int dim_scale = 0;
for (unsigned int d = 0; d < dim; ++d)
  dim_scale += (long(floor(p[d]*100)) % 2);

return (dim_scale == dim ? 100. : 10.);
```

Rombur: sure

Rombur:
ok done

PR #75, opened 2018-06-01
Add matrix-free laplace code and test

Rombur:
This is the matrix-free equivalent of the `laplace` code

### DIFF
--- a/tests/test_laplace_matrix_free.cc
+++ b/tests/test_laplace_matrix_free.cc
@@ -1,0 +1,133 @@
+/*************************************************************************
+ * Copyright (c) 2017-2018 by the mfmg authors                           *
+ * All rights reserved.                                                  *
+ *                                                                       *
+ * This file is part of the mfmg libary. mfmg is distributed under a BSD *
+ * 3-clause license. For the licensing terms see the LICENSE file in the *
+ * top-level directory                                                   *
+ *                                                                       *
+ * SPDX-License-Identifier: BSD-3-Clause                                 *
+ *************************************************************************/
+
+#define BOOST_TEST_MODULE laplace_matrix_free
+
+#include "laplace_matrix_free.hpp"
+#include "main.cc"
+
+#include <deal.II/lac/precondition.h>
+
+namespace tt = boost::test_tools;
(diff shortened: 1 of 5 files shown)


tests/test_laplace_matrix_free.cc
```diff
@@ -0,0 +1,133 @@
...
+  Source<dim> source;
+
+  Laplace<dim, fe_degree, double> laplace(MPI_COMM_WORLD);
+  laplace.setup_system(boost::property_tree::ptree(), material_property);
+  laplace.assemble_rhs(source);
+
+  dealii::PreconditionIdentity preconditioner;
+  laplace.solve(preconditioner);
+
+  // The exact solution is quadratique so the error should be zero.
```
aprokop: No French! :)

tests/test_laplace_matrix_free.cc
```diff
@@ -0,0 +1,133 @@
...
```
aprokop: French

tests/test_laplace_matrix_free.cc
Rombur: That's the fancy spelling, it adds a  je ne sais quoi  to the library.

Rombur:
>  1.   It seems that one cannot include both laplace.hpp and laplace_matrix_free.hpp due to clashing Laplace class names.

Sure, I will change the names

>  2.   There seems to be a lot of commonalities between two implementations of Laplace. Do you think it would be hard to merge them together with a matrix-free being a parameter option?

That's what I thought at first too but when you compare the code more closely there is actually little overlap except to create the geometry

aprokop:
:( Please do not rebase PRs until approved.

Rombur:
But you approved the PR! I was going to merge myself...

aprokop:
Lol, I forgot about it, and wanted to check the changes. Mea culpa.

aprokop:
Why did you need to update the tolerance?

Rombur:
The tester was not working since we updated the docker images. The build was crashing during the configuration but jenkins reported that everything was fine....

PR #77, opened 2018-06-21
Add supports for zoltan and metis to build agglomerates

Rombur:
This adds the possibility to use zoltan or metis to create the agglomerate. Since zoltan and metis are called locally on each processor, agglomerates cannot cross processor boundary. This PR adds a new function `build_agglomerates_partitioner` and move `build_agglomerates` to `build_agglomerates_block`.

### DIFF
--- a/include/mfmg/amge.templates.hpp
+++ b/include/mfmg/amge.templates.hpp
@@ -16,12 +16,15 @@
 
 #include <mfmg/exceptions.hpp>
 
+#include <deal.II/distributed/tria.h>
 #include <deal.II/dofs/dof_accessor.h>
 #include <deal.II/grid/grid_tools.h>
+#include <deal.II/lac/sparsity_tools.h>
 #include <deal.II/numerics/data_out.h>
 
 #include <algorithm>
 #include <fstream>
+#include <map>
 
 namespace mfmg
 {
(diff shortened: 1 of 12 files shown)


include/mfmg/amge.templates.hpp
```diff
@@ -34,89 +37,33 @@ AMGe<dim, VectorType>::AMGe(MPI_Comm comm,
...
-  unsigned int constexpr x_p = 1;
-  unsigned int constexpr y_p = 3;
-  unsigned int constexpr z_p = 5;
-
-  // Flag the cells to create the agglomerates
-  unsigned int agglomerate = 1;
-  for (auto cell : _dof_handler.active_cell_iterators())
+  std::string partitioner_type = ptree.get<std::string>("partitioner");
+  std::transform(partitioner_type.begin(), partitioner_type.end(),
+                 partitioner_type.begin(), ::tolower);
```
dalg24: `std::tolower` from `<locale>`?

include/mfmg/amge.templates.hpp
```diff
@@ -302,6 +249,171 @@ void AMGe<dim, VectorType>::compute_restriction_sparse_matrix(
...
+        }
+        else
+          break;
+      }
+
+      ++agglomerate;
+    }
+  }
+
+  return agglomerate - 1;
```
dalg24: Is that because you meant to sent non-zero user indices?

include/mfmg/amge.templates.hpp
Rombur: no it's a C function but it's similar

include/mfmg/amge.templates.hpp
Rombur: Yes, I just wanted an easy way to check that the cells were flagged correctly that's why I didn't want to use zero which is the default value of the user index

Rombur:
How cruel of you to let a PR alone and afraid! Will someone have pity and merge it?

Rombur:
Thanks!

PR #79, opened 2018-06-29
Implement alternative approaches to compute eigenvectors with ARPACK

dalg24:


Rombur:
You need to change the default approach or your CG parameter if you want the tests to pass

PR #80, opened 2018-08-28
Dealii dev

Rombur:
Update the docker image and mfmg to use the development version of deal. The reason for this change is that it fixes the incompatibility between deal.II and trilinos. Right now, you need to be lucky when you pick a version of Trilinos. This has been "fixed" in deal by making sacado an optional dependency.

This also fixes the bug found by @dalg24 when using `mmult`.

 Note that you will need to update your version of deal.II

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,6 +1,6 @@
 FROM nvidia/cuda:8.0-devel
 
-ENV N_PROCS=4
+ENV N_PROCS=1
 
 RUN apt-get update && apt-get upgrade -y && apt-get install -y \
       gcc \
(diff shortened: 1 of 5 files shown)


ci/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM nvidia/cuda:8.0-devel
 
-ENV N_PROCS=4
+ENV N_PROCS=1
```
dalg24: Why did you change the default?

ci/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN mkdir -p ${PREFIX} && \
     mkdir source && \
     mkdir build
 
-RUN export CMAKE_VERSION=3.11.1 && \
+RUN export CMAKE_VERSION=3.11.4 && \
```
dalg24: What version of CMake do we actually require?  Just curious.

ci/Dockerfile
Rombur: Because I can't compile with 4 procs on my desktop. deal uses all the memory.

PR #81, opened 2018-08-28
Fix bug with Zoltan

Rombur:
The first commit updates the documentation that I forgot to do in the previous PR.

The second commit fixes a bug when using Zoltan to create agglomerates. The problem is that Zoltan was creating agglomerates with IDs between 0 and n-1 but we have reserved ID for artificial and ghost cells. So now we renumber the agglomerates from 1 to n.


doc/install.rst
```diff
@@ -8,33 +8,26 @@ Install third-party libraries
...
++------------------------+------------------------+
+| Boost                  | 1.65.1                 |
++------------------------+------------------------+
+| BLAS/LAPACK            | N/A                    |
++------------------------+------------------------+
+| deal.II                | development            |
++------------------------+------------------------+
+| MPI                    | N/A                    |
++------------------------+------------------------+
+| Trilinos               | 12.X                   |
```
dalg24: `12.12.1`?

include/mfmg/amge.templates.hpp
```diff
@@ -401,17 +401,22 @@ unsigned int AMGe<dim, VectorType>::build_agglomerates_partitioner(
...
   for (auto cell : _dof_handler.active_cell_iterators())
   {
     if (cell->is_locally_owned())
     {
       unsigned int const index =
           index_map.at(std::make_pair(cell->level(), cell->index()));
-      cell->set_user_index(partition_indices[index]);
+      cell->set_user_index(partition_indices[index] + 1);
+      if (n_zoltan_agglomerates < partition_indices[index] + 1)
+        n_zoltan_agglomerates = partition_indices[index] + 1;
```
dalg24: Why did you prefer this over `return *std::max_element(partition_indices.begin(), partition_indices.end()) + 1;`?

doc/install.rst
Rombur: We only use `Epetra` and `EpetraExt` which haven't changed in ages so any version of Trilinos should be good.

include/mfmg/amge.templates.hpp
Rombur: Since I am already going through `partition_indices`, I thought that I could save the state at the same time.

Rombur:
Can I merge this?

PR #82, opened 2018-08-28
Improve CMake

Rombur:
This PR allows to build the mfmg as a shared library even when CUDA is enabled.

@aprokop can you try this PR on your machine


aprokop:
Sure, once I get dealii installed :)

Rombur:
@aprokop I know this does not pass the tests but can you still try it. I would be interested to see if the problem is reproducible outside docker.

Rombur:
I am giving up on this for now. While removing the stubs fixes the problem in docker it seems that stubs are not the real reason of the problem.

PR #83, opened 2018-08-30
Fix a bug with zoltan where agglomerate IDs where not consecutive

Rombur:
The agglomerate IDs created by Zoltan are not always consecutive. This causes the code to crash inside deal when we try to create the Triangulation of an agglomerate that does not exist. Fix that bug by assigning our own IDs to the agglomerates.


include/mfmg/amge.templates.hpp
```diff
@@ -401,18 +402,27 @@ unsigned int AMGe<dim, VectorType>::build_agglomerates_partitioner(
   dealii::SparsityTools::partition(cell_connectivity, n_agglomerates,
                                    partition_indices, partitioner);
 
-  // Assign the agglomerate ID to all the locally owned cells. Add one to the
-  // agglomerate ID because zero is reserved for the ghost and artificial cells.
+  // Assign the agglomerate ID to all the locally owned cells. Zoltan does not
+  // guarantee that the agglomerate IDs will consecutive so we need to
+  // renumbered them.
```
dalg24: s/renumbered/renumber/

include/mfmg/amge.templates.hpp
```diff
@@ -401,18 +402,27 @@ unsigned int AMGe<dim, VectorType>::build_agglomerates_partitioner(
...
     {
       unsigned int const index =
           index_map.at(std::make_pair(cell->level(), cell->index()));
-      cell->set_user_index(partition_indices[index] + 1);
-      if (n_zoltan_agglomerates < partition_indices[index] + 1)
-        n_zoltan_agglomerates = partition_indices[index] + 1;
+      auto agg_id = agglomerate_renumbering.find(partition_indices[index]);
+      if (agg_id == agglomerate_renumbering.end())
+      {
+        ++n_zoltan_agglomerates;
```
dalg24: might not be a bad idea to keep a comment that zero is reserved for the ghost and artificial cells and that's why you map indices from 1 to n.

Rombur: Done. I have added the comment above the for loop.

dalg24:
Ready to merge?

Rombur:
yes

PR #84, opened 2018-09-07
Only require action of the operator on a vector

dalg24:
A_c = R A_f P with P = R^T

Now we only need the action of A_f on rows of the restrictor.


include/mfmg/dealii_operator_device.cuh
```diff
@@ -51,6 +51,12 @@ public:
   virtual std::shared_ptr<operator_type>
   multiply(operator_type const &operator_b) const override final;
 
+  std::shared_ptr<operator_type>
+  multiply_transpose(operator_type const &operator_b) const
+  {
+    return std::make_shared<SparseMatrixDeviceOperator<VectorType>>(nullptr);
```
dalg24: Of course `test_hierarchy_device` does not like that one too much...

include/mfmg/dealii_operator.templates.hpp
```diff
@@ -148,6 +148,146 @@ DealIITrilinosMatrixOperator<VectorType>::build_range_vector() const
...
+  auto downcast_operator_b =
+      static_cast<DealIITrilinosMatrixOperator<VectorType> const &>(operator_b);
+
+  auto a = this->get_matrix();
+  auto b = downcast_operator_b.get_matrix();
+
+  auto c = std::make_shared<matrix_type>(a->locally_owned_range_indices(),
+                                         b->locally_owned_range_indices(),
+                                         a->get_mpi_communicator());
+  // a->mmult(*c, *b);
```
dalg24: Remove commented line.

include/mfmg/dealii_operator.templates.hpp
```diff
@@ -148,6 +148,146 @@ DealIITrilinosMatrixOperator<VectorType>::build_range_vector() const
...
+
+  matrix_transpose_matrix_multiply(*c, *b, *a);
+
+  return std::make_shared<DealIITrilinosMatrixfreeOperator<VectorType>>(c);
+}
+
+template <typename VectorType>
+std::shared_ptr<MatrixOperator<VectorType>>
+DealIITrilinosMatrixfreeOperator<VectorType>::multiply(
+    MatrixOperator<VectorType> const &operator_b) const
```
dalg24: Actually not sure that `multiply()` should be defined in `DealIITrilinosMatrixfreeOperator`.  I needed it at some point when I was doing changes by increments...  If we remove it, let's not forget to put `final` specifier back for `DealIITrilinosMatrixOperator::multiply()`.

include/mfmg/dealii_operator.templates.hpp
```diff
@@ -148,6 +148,96 @@ DealIITrilinosMatrixOperator<VectorType>::build_range_vector() const
...
+  std::vector<dealii::types::global_dof_index> i_indices;
+  A.locally_owned_range_indices().fill_index_vector(i_indices);
+  std::vector<dealii::types::global_dof_index> j_indices;
+  B.locally_owned_range_indices().fill_index_vector(j_indices);
+
+  int const global_n_rows = A.row_partitioner().NumGlobalElements();
+  int const global_n_columns = B.row_partitioner().NumGlobalElements();
+  for (int j = 0; j < global_n_columns; ++j)
+  {
+    auto const local_j = B.locally_owned_range_indices().index_within_set(j);
```
Rombur: This is not used.

include/mfmg/dealii_operator.templates.hpp
```diff
@@ -148,6 +148,96 @@ DealIITrilinosMatrixOperator<VectorType>::build_range_vector() const
...
+    auto const error_code = matrix.trilinos_matrix().ExtractMyRowView(
+        local_j, num_entries, values, local_indices);
+    ASSERT(error_code == 0,
+           "Non-zero error code (" + std::to_string(error_code) +
+               ") returned by Epetra_CrsMatrix::ExtractMyRowView()");
+  }
+
+  dealii::IndexSet ghost_indices(matrix.locally_owned_domain_indices().size());
+  for (int k = 0; k < num_entries; ++k)
+    ghost_indices.add_index(matrix.trilinos_matrix().GCID(local_indices[k]));
```
Rombur: This creates a "custom" `IndexSet` for this row, which is different than the global `Epetra_Map`, right?

include/mfmg/dealii_operator.templates.hpp
Rombur: If it's not necessary, let's remove it.

include/mfmg/dealii_operator.hpp
```diff
@@ -129,6 +129,9 @@ class DealIITrilinosMatrixfreeOperator
       std::shared_ptr<matrix_type> matrix,
       std::shared_ptr<sparsity_pattern_type> sparsity_pattern = nullptr);
 
+  std::shared_ptr<operator_type>
+  multiply_transpose(operator_type const &operator_b) const;
```
Rombur: Add a comment that we are doing `*this * operator_b^T`

include/mfmg/dealii_adapters.hpp
```diff
@@ -36,7 +36,7 @@ class DealIIMeshEvaluator
   using vector_type = VectorType;
   using value_type = typename VectorType::value_type;
   using operator_type = Operator<vector_type>;
-  using global_operator_type = DealIITrilinosMatrixOperator<vector_type>;
+  using global_operator_type = DealIITrilinosMatrixfreeOperator<vector_type>;
```
Rombur: This means that we cannot use the matrix-based without recompiling the code and thus, we cannot test the matrix-based code. Can we instead derive `DealIIMatrixFreeMeshEvaluator` from `DealIIMeshEvaluator`? We will need to add a specialization of `Adapter` though. I think that we should have used inheritance instead of template specialization for `Adapted` but that's for another PR.

tests/test_hierarchy.cc
```diff
@@ -220,3 +223,118 @@ BOOST_AUTO_TEST_CASE(zoltan)
     BOOST_TEST(conv_rate == ref_solution, tt::tolerance(1e-6));
   }
 }
+
+dealii::TrilinosWrappers::SparseMatrix
+gimme_a_matrix(unsigned int n_local_rows, unsigned int n_entries_per_row)
```
Rombur: ...

tests/test_hierarchy.cc
```diff
@@ -220,3 +223,118 @@ BOOST_AUTO_TEST_CASE(zoltan)
...
+      column_indices.insert(column_index);
+    }
+    nnz += column_indices.size();
+  }
+  sparse_matrix.compress(dealii::VectorOperation::insert);
+  return sparse_matrix;
+}
+
+dealii::TrilinosWrappers::SparseMatrix
+gimme_identity(unsigned int n_local_rows = 10)
```
Rombur: Why does this function has a default argument?

tests/test_hierarchy.cc
```diff
@@ -220,3 +223,118 @@ BOOST_AUTO_TEST_CASE(zoltan)
...
 }
+
+dealii::TrilinosWrappers::SparseMatrix
+gimme_a_matrix(unsigned int n_local_rows, unsigned int n_entries_per_row)
+{
+  unsigned int const comm_size =
+      dealii::Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
+  unsigned int const comm_rank =
+      dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
+  unsigned int const size = comm_size * n_local_rows;
```
Rombur: This only works if `n_local_rows` is the same on every processor, which is fine, but add a comment to make it clear.

tests/test_hierarchy.cc
```diff
@@ -220,3 +223,118 @@ BOOST_AUTO_TEST_CASE(zoltan)
...
+
+  unsigned int nnz = 0;
+  std::default_random_engine generator;
+  for (unsigned int i = 0; i < n_local_rows; ++i)
+  {
+    std::uniform_int_distribution<int> distribution(0, size - 1);
+    std::set<int> column_indices;
+    for (unsigned int j = 0; j < n_entries_per_row; ++j)
+    {
+      int column_index = distribution(generator); // FIXME
```
Rombur: What's the `FIXME` for?

include/mfmg/dealii_operator.hpp
```diff
@@ -114,6 +119,25 @@ class DealIITrilinosMatrixOperator : public MatrixOperator<VectorType>
   std::shared_ptr<matrix_type> _matrix;
 };
 
+template <typename VectorType>
+class DealIITrilinosMatrixfreeOperator
```
Rombur: A better name for this would be `DealIIMatrixFreeOperator` because ideally we won't have any matrices except at the coarsest level.

tests/test_hierarchy.cc
```diff
@@ -220,3 +223,118 @@ BOOST_AUTO_TEST_CASE(zoltan)
...
+      A.locally_owned_domain_indices(), A.get_mpi_communicator());
+
+  C.vmult(v, u);
+  BOOST_TEST(v.l2_norm() > 1.);
+
+  u = -1.;
+  C_ref.vmult_add(v, u);
+  BOOST_TEST(v.l2_norm() == 0.);
+
+  // unsigned int const comm_rank =
```
Rombur: Do you want to keep this?

PR #87, opened 2018-10-17
Correctly install header files

Rombur:


CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ IF(${MFMG_ENABLE_DOCUMENTATION})
 ENDIF()
 
 INCLUDE_DIRECTORIES(include)
-FILE(GLOB MFMG_HEADERS include/*.hpp)
+FILE(GLOB MFMG_HEADERS "include/*.hpp" "include/*.cuh")
```
aprokop: May want to use `CONFIGURE_DEPENDS` for globbing (assuming CMake is recent enough).
Can also remove the internal quotes, and just use `"include/*.hpp include/*.cuh"`, I think

Rombur: `CONFIGURE_DEPENDS` is only in cmake 3.12 which we do not support.

PR #89, opened 2018-10-17
Update arpack patch for newer version of deal.II

Rombur:


aprokop:
It's a bit unclear why this patch is coming without any updates to the docker container.

Rombur:
Because I forgot to update patch when I updated the image.

aprokop:
Just trying to figure things out... So you used this updated patch when creating the new docker image, but then just forgot to include it with image update, correct?

Rombur:
@aprokop No, I should have but I forgot. This patch allows to use arpack in a way that will be useful for matrix-free but right now we don't have any test.

aprokop:
Ah, thank you. I was wondering why we did not detect the broken docker image. No test completely answers this.

Rombur:
> No test completely answers this.

That's how we do real software development. If we don't test for it, we cannot break the testsuite!

PR #96, opened 2018-11-02
Implement MatrixFreeHierarchyHelpers

### LINKED ISSUE #91: MatrixFreeHierarchyHelpers
We need to implement  `MatrixFreeHierarchyHelpers` to use the matrix-free implementation the easiest will probably to derive it (for now) frm `DealIIHierarchyHelpers`.

dalg24:
Closes #91 

~~Drop e9116f5 before merging the changes.~~ Done!
~~Probably should define another evaluator type name for it and test both for the legacy and the new matrix free types.~~ Added `mesh_evaluator_type` string argument with default value `"DealIIMeshEvaluator"` instead

### DIFF
--- a/include/mfmg/common/hierarchy.hpp
+++ b/include/mfmg/common/hierarchy.hpp
@@ -45,6 +45,18 @@ create_hierarchy_helpers(std::shared_ptr<MeshEvaluator const> evaluator)
     else
       ASSERT_THROW_NOT_IMPLEMENTED();
   }
+  else if (evaluator_type == "DealIIMatrixFreeMeshEvaluator")
+  {
+    int const dim = evaluator->get_dim();
+    if (dim == 2)
+      hierarchy_helpers.reset(
+          new DealIIMatrixFreeHierarchyHelpers<2, VectorType>());
+    else if (dim == 3)
+      hierarchy_helpers.reset(
+          new DealIIMatrixFreeHierarchyHelpers<3, VectorType>());
+    else
+      ASSERT_THROW_NOT_IMPLEMENTED();
+  }
   else if (evaluator_type == "CudaMeshEvaluator")
   {
     int const dim = evaluator->get_dim();
--- a/include/mfmg/common/hierarchy.hpp
+++ b/include/mfmg/common/hierarchy.hpp
@@ -20,6 +20,7 @@
 #include <mfmg/cuda/cuda_matrix_operator.cuh>
 #include <mfmg/cuda/cuda_mesh_evaluator.cuh>
 #include <mfmg/dealii/dealii_hierarchy_helpers.hpp>
+#include <mfmg/dealii/dealii_matrix_free_hierarchy_helpers.hpp>
 
 #include <boost/property_tree/ptree.hpp>
 
--- a/include/mfmg/common/operator.hpp
+++ b/include/mfmg/common/operator.hpp
@@ -23,6 +23,8 @@ public:
   using operator_type = Operator<VectorType>;
   using vector_type = VectorType;
 
+  virtual ~Operator() = default;
+
   virtual void apply(vector_type const &x, vector_type &y) const = 0;
 
   virtual std::shared_ptr<operator_type> transpose() const = 0;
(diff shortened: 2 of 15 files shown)


source/dealii/dealii_matrix_free_hierarchy_helpers.cc
```diff
@@ -0,0 +1,92 @@
...
+
+  auto restrictor_matrix =
+      std::make_shared<dealii::TrilinosWrappers::SparseMatrix>();
+  auto global_operator = get_global_operator(mesh_evaluator);
+  auto matrix_free_global_operator =
+      std::dynamic_pointer_cast<DealIIMatrixFreeOperator<VectorType>>(
+          global_operator);
+  // ughhh
+  auto system_sparse_matrix = matrix_free_global_operator->get_matrix();
+  // why do we pass system matrix to amge?
```
dalg24: Need to discuss this with Bruno

source/dealii/dealii_matrix_free_hierarchy_helpers.cc
```diff
@@ -0,0 +1,92 @@
...
+        dealii_mesh_evaluator->get_constraints(), *system_matrix);
+
+    this->_global_operator.reset(
+        new DealIIMatrixFreeOperator<VectorType>(system_matrix));
+  }
+
+  return this->_global_operator;
+}
+
+// copy/paste from DealIIHierarchyHelpers::build_restrictor()
```
dalg24: Don't forget to update that comment.

include/mfmg/common/hierarchy.hpp
```diff
@@ -38,9 +39,11 @@ create_hierarchy_helpers(std::shared_ptr<MeshEvaluator const> evaluator)
   {
     int const dim = evaluator->get_dim();
     if (dim == 2)
-      hierarchy_helpers.reset(new DealIIHierarchyHelpers<2, VectorType>());
+      hierarchy_helpers.reset(
+          new DealIIMatrixFreeHierarchyHelpers<2, VectorType>());
     else if (dim == 3)
-      hierarchy_helpers.reset(new DealIIHierarchyHelpers<3, VectorType>());
+      hierarchy_helpers.reset(
+          new DealIIMatrixFreeHierarchyHelpers<3, VectorType>());
```
dalg24: DO NOT MERGE THESE CHANGES AS IS

include/mfmg/common/operator.hpp
```diff
@@ -23,6 +23,8 @@ class Operator
   using operator_type = Operator<VectorType>;
   using vector_type = VectorType;
 
+  virtual ~Operator() = default;
```
dalg24: Make sure Bruno sees this.  If necessary make a separate PR for it.

dalg24:
I don't see how to do this without creating a new class deriving from `DealIIMeshEvaluator` which sounds a bit silly to me.

Rombur:
We could add an input parameter to `DealIIMeshEvaluator` which could then return `DealIIMeshEvaluator` or `DealIIMatrixFreeMeshEvaluator`. Not the most elegant but we need a way to switch the implementation without having to recompile the code everytime.

dalg24:
> We could add an input parameter to `DealIIMeshEvaluator` which could then return `DealIIMeshEvaluator` or `DealIIMatrixFreeMeshEvaluator`. Not the most elegant but we need a way to switch the implementation without having to recompile the code everytime.

Fine by me

dalg24:
Zoltan unit test in test_hierarchy does not seems to like the matrix free implementation.

Rombur:
> Zoltan unit test in test_hierarchy does not seems to like the matrix free implementation.

Yeah, strange. The unit test is doesn't do anything special it just uses a different partition of the mesh. It is just less regular than the other partitions.

Rombur:
So the reason why the Zoltan test fails is a little bit strange. The second time that we call Zoltan we get a different partitioning than the first time. I tried to split the single test into two but it didn't change anything. It looks like Zoltan uses some variables that are not reset correctly between runs. I looked at deal.II's code but everything looks fine. Let's just run this test for a single case and write a comment about this problem with Zoltan.

source/dealii/dealii_matrix_free_hierarchy_helpers.cc
Rombur: See #97

PR #105, opened 2018-12-06
Remove explicit prolongator construction

aprokop:
This PR adds support for `apply()` in transpose mode, and removes prolongator from the hierarchy, replacing it by action of restrictor in transpose mode.

As we don't have a proper support for transpose action in device sparse matrices, I chose to workaround that by using the implemented `transpose()` call of `CudaMatrixOperator`. That class now caches the matrix transpose when needed. I am not sure how to avoid `mutable` class variables, though, and open to suggestions.

### DIFF
--- a/include/mfmg/common/level.hpp
+++ b/include/mfmg/common/level.hpp
@@ -36,11 +36,6 @@ public:
     return _restrictor;
   }
 
-  std::shared_ptr<operator_type const> get_prolongator() const
-  {
-    return _prolongator;
-  }
-
   std::shared_ptr<Smoother<vector_type> const> get_smoother() const
   {
     return _smoother;
@@ -58,11 +53,6 @@ public:
     _restrictor = r;
   }
 
-  void set_prolongator(std::shared_ptr<operator_type const> p)
-  {
-    _prolongator = p;
-  }
-
   void set_smoother(std::shared_ptr<Smoother<vector_type> const> s)
   {
     _smoother = s;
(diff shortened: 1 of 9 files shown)


include/mfmg/common/level.hpp
```diff
@@ -58,20 +53,22 @@ class Level
...
   {
     _smoother = s;
   }
 
   void set_solver(std::shared_ptr<Solver<vector_type> const> s) { _solver = s; }
 
+  std::shared_ptr<vector_type> build_vector() const
+  {
+    auto a = get_operator();
+    assert(a);
```
Rombur: We have our own assert

include/mfmg/common/operator.hpp
```diff
@@ -16,6 +16,12 @@
 
 namespace mfmg
 {
+enum OperatorMode
```
Rombur: Use `enum class`

include/mfmg/cuda/cuda_matrix_operator.cuh
```diff
@@ -50,7 +51,8 @@ public:
   std::shared_ptr<SparseMatrixDevice<value_type>> get_matrix() const;
 
 private:
-  std::shared_ptr<SparseMatrixDevice<value_type>> _matrix;
+  mutable std::shared_ptr<SparseMatrixDevice<value_type>> _matrix;
+  mutable std::shared_ptr<SparseMatrixDevice<value_type>> _transposed_matrix;
```
Rombur: Why do you need to store both matrices?

include/mfmg/common/operator.hpp
aprokop: Thank you, was not aware of this.

include/mfmg/cuda/cuda_matrix_operator.cuh
aprokop: The goal was to try to minimize the number of transpose calculations. Transpose actually appears in multiple places (through `apply(TRANS)` and `transpose()`, for example), so by storing it we essentially cache the calculation.

PR #108, opened 2018-12-12
Update docker and require C++14

Rombur:
This PR updates the docker image and requires C++14 instead of C++11. The reason I didn't do it before is because my desktop was using cuda 8.0 but I have switched to cuda 10. @dalg24 @aprokop does this work for the both of you if we drop support cuda 8.0? In this PR jenkins uses cuda 9.2 because we don't have the drivers on fetnat for cuda 10.


dalg24:
Did you already build the image and pushed it to docker hub?

Rombur:
> Did you already build the image and pushed it to docker hub?
Yeah I am not sure why so many tests timed out.

Rombur:
I have remade the image. Let's give it another shot.

Rombur:
Are you fine merging this? If yes, I will retag the image as latest and then merge it.

ci/Dockerfile
```diff
@@ -223,7 +223,8 @@ RUN export DEAL_II_HASH=1a364b58c4a5e2235c1282544d55486d41bead96 && \
         -DDEAL_II_WITH_ARPACK=ON \
         -DARPACK_DIR=${ARPACK_DIR} \
         -DDEAL_II_WITH_CUDA=ON \
-        -DDEAL_II_WITH_CXX14=OFF \
+        -DDEAL_II_WITH_CXX14=ON \
+        -DDEAL_II_WITH_CXX17=OFF \
```
dalg24: Do we actually need the `DEAL_II_WITH_CXX17=OFF`?

dalg24:
Go ahead

ci/Dockerfile
Rombur: It should be off by default but this way it is obvious that it is. No version of cuda support c++17.

PR #110, opened 2018-12-14
Use boundary on the agglomerates

Rombur:
This PR add support for boundary on the agglomerates. With it, I can reproduce the result from the paper. Note that the convergence rate is sometimes better and sometimes worse. Also, I had to decrease the randomness of the mesh because the coarse matrix was too ill-conditioned for the direct solver. Trilinos would throw an error but scipy worked.

### DIFF
--- a/include/mfmg/common/amge.templates.hpp
+++ b/include/mfmg/common/amge.templates.hpp
@@ -91,16 +91,52 @@ void AMGe<dim, VectorType>::build_agglomerate_triangulation(
 {
   std::vector<typename dealii::DoFHandler<dim>::active_cell_iterator>
       agglomerate;
+  // Map between the cells on the boundary and the faces on the boundary and the
+  // associated boundary id.
+  std::map<typename dealii::DoFHandler<dim>::active_cell_iterator,
+           std::vector<std::pair<unsigned int, unsigned int>>>
+      boundary_ids;
   for (auto cell : _dof_handler.active_cell_iterators())
     if (cell->user_index() == agglomerate_id)
+    {
       agglomerate.push_back(cell);
+      if (cell->at_boundary())
+      {
+        for (unsigned int f = 0; f < dealii::GeometryInfo<dim>::faces_per_cell;
+             ++f)
+        {
+          if (cell->face(f)->at_boundary())
+          {
+            boundary_ids[cell].push_back(
+                std::make_pair(f, cell->face(f)->boundary_id()));
+          }
+        }
+      }
+    }
 
   // If the agglomerate has hanging nodes, the patch is bigger than
-  // what we may expect because we cannot a create a coarse triangulation with
+  // what we may expect because we cannot create a coarse triangulation with
   // hanging nodes. Thus, we need to use FE_Nothing to get ride of unwanted
   // cells.
   dealii::GridTools::build_triangulation_from_patch<dealii::DoFHandler<dim>>(
       agglomerate, agglomerate_triangulation, agglomerate_to_global_tria_map);
+
+  // Copy the boundary IDs to the agglomerate triangulation
+  for (auto const &boundary : boundary_ids)
+  {
+    auto const boundary_cell = boundary.first;
+    for (auto &agglomerate_cell : agglomerate_to_global_tria_map)
+    {
+      if (agglomerate_cell.second == boundary_cell)
+      {
+        for (auto &boundary_face : boundary.second)
+          agglomerate_cell.first->face(boundary_face.first)
+              ->set_boundary_id(boundary_face.second);
+
+        break;
+      }
+    }
+  }
 }
 
 template <int dim, typename VectorType>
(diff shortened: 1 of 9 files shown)


aprokop:
sorry for the delay, will  look at it now.

include/mfmg/common/amge.templates.hpp
```diff
@@ -91,16 +91,50 @@ void AMGe<dim, VectorType>::build_agglomerate_triangulation(
 {
   std::vector<typename dealii::DoFHandler<dim>::active_cell_iterator>
       agglomerate;
+  std::map<typename dealii::DoFHandler<dim>::active_cell_iterator,
```
aprokop: Add a comment describing what maps to what, "cell iterator -> (face index in cell, boundary id)".

include/mfmg/common/amge.templates.hpp
```diff
@@ -231,6 +265,44 @@ void AMGe<dim, VectorType>::compute_restriction_sparse_matrix(
...
 #if MFMG_DEBUG
+  // Check that the locally_relevant_global_diag is the sum of the agglomerates
+  // diagonal
+  // TODO do not ask user for the locally_relevant_global_diag
+  pos = 0;
+  dealii::LinearAlgebra::distributed::Vector<typename VectorType::value_type>
+      new_global_diag(locally_relevant_global_diag.get_partitioner());
+  for (unsigned int i = 0; i < n_agglomerates; ++i)
+  {
+    unsigned int const n_local_eig = n_local_eigenvectors[i];
```
aprokop: Could you avoid using `n_local_eigenvetors` here and `eigenvectors` in the loop below? It makes it seem as if you are using some eigenvectors when computing the diagonal, when in fact you just get the right sizes out. Would be easier to read that if you just get those sizes out some other way.

include/mfmg/common/amge.templates.hpp
```diff
@@ -231,6 +265,44 @@ void AMGe<dim, VectorType>::compute_restriction_sparse_matrix(
...
+              dof_indices_maps[i][j];
+          new_global_diag[global_pos] += diag_elements[i][j];
+        }
+      }
+      ++pos;
+    }
+  }
+  new_global_diag.compress(dealii::VectorOperation::add);
+  new_global_diag -= locally_relevant_global_diag;
+  ASSERT((new_global_diag.l2_norm() / locally_relevant_global_diag.l2_norm()) <
```
aprokop: You probably don't want `l2` norm here, and want some `max` norm. Ideally, each diagonal element checked individually in relative norm.

include/mfmg/common/amge.templates.hpp
```diff
@@ -231,6 +265,44 @@ void AMGe<dim, VectorType>::compute_restriction_sparse_matrix(
...
+             1e-14,
+         "Sum of agglomerate diagonals is not equal to the global diagonal");
+
+  //  new_global_diag.print(std::cout);
```
aprokop: Remove `cout` comments.

aprokop:
I don't quite understand where ill-conditioness is coming from. What does 0.2 fraction mesh distortion correspond to? If it's a fraction of the distance between points, then that should produce a mesh that is still pretty well conditioned.

Rombur:
> What does 0.2 fraction mesh distortion correspond to? If it's a fraction of the distance between points, then that should produce a mesh that is still pretty well conditioned.

Yes, that's what it is. I am not sure why there is a problem. This only happens when the domain is a cube, when the domain is a ball there is no problem.

aprokop:
What does direct solver say? Is it possible that the problem is coming from the singularity of the matrix (and, thus, possibly singularity of the coarse matrix) instead? This is typically the problem with Neumann b.c.s, though we use Dirichlet here.

Rombur:
It crashes with error -22. I couldn't find what it means. I still have the coarse matrix if you are interested. It's a sparse 15 by 15 matrix. You can load it in python and look at it. It is symmetric but not definite positive.

aprokop:
Sure, could you attach it to this issue? And what was the direct solver?

Rombur:
Here is the [matrix](https://github.com/ORNL-CEES/mfmg/files/2687806/mat_4.txt)
Here is a python code to load the matrix
```python
col = []
row = []
val = []
with open('mat_4.txt', 'r') as csvfile:
    csvfile = csv.reader(csvfile, delimiter=' ')
    for r in csvfile:
        row.append(int(r[0]))
        col.append(int(r[1]))
        val.append(float(r[2]))
mat = scipy.sparse.coo_matrix((val,(row,col)))
```
The direct solver is KLU

Rombur:
Done but I need to squash the `fixup` before this is merged.

Rombur:
This is ready for a new round of review. I have also added a commit that make the laplace test more robust. I still need to squash a few commits before this can be merged.

include/mfmg/common/amge.templates.hpp
```diff
@@ -231,6 +267,34 @@ void AMGe<dim, VectorType>::compute_restriction_sparse_matrix(
...
+  }
+  new_global_diag.compress(dealii::VectorOperation::add);
+  new_global_diag -= locally_relevant_global_diag;
+  ASSERT((new_global_diag.linfty_norm() /
+          locally_relevant_global_diag.linfty_norm()) < 1e-14,
+         "Sum of agglomerate diagonals is not equal to the global diagonal");
+
+  dealii::LinearAlgebra::distributed::Vector<typename VectorType::value_type>
+      diff = locally_relevant_global_diag;
+  diff -= new_global_diag;
```
dalg24: What is the point of `diff`?

tests/laplace.hpp
```diff
@@ -96,8 +96,20 @@ void Laplace<dim, VectorType>::setup_system(
...
+  for (auto &cell : boundary_cells)
+  {
+    for (unsigned int f = 0; f < dealii::GeometryInfo<dim>::faces_per_cell; ++f)
+      if (cell->face(f)->at_boundary())
+        cell->face(f)->set_boundary_id(1);
+  }
+
   if (ptree.get("distort_random", false))
-    dealii::GridTools::distort_random(0.2, _triangulation);
+    dealii::GridTools::distort_random(0.1, _triangulation);
```
dalg24: I was going to ask about this change but I saw the commit message.

tests/test_restriction_matrix.cc
```diff
@@ -230,6 +230,9 @@ class TestMeshEvaluator : public mfmg::DealIIMeshEvaluator<dim>
     constraints.clear();
     constraints.reinit(locally_relevant_dofs);
     dealii::DoFTools::make_hanging_node_constraints(dof_handler, constraints);
+    // The test requires us not to put boundary conditions
+    // dealii::VectorTools::interpolate_boundary_values(
+    //     dof_handler, 1, dealii::Functions::ZeroFunction<dim>(), constraints);
```
dalg24: Can you elaborate?  I don't really understand.

include/mfmg/common/amge.templates.hpp
Rombur: Left over code. It shouldn't be there.

tests/test_restriction_matrix.cc
Rombur: The test assumes that the eigenvectors are constant. This is true for laplace if you don't have dirichlet bc. If you have dirichlet bc, the eigenvectors are not constant and the test fails.

Rombur:
@aprokop can we merge this?

Rombur:
_ping_

Rombur:
I have talked to @aprokop and we agreed to merge this PR as is.

PR #121, opened 2019-01-24
Lanczos

wdj:
See test_lanczos.cc for example of use; that code does one simple test.

An issue that may need addressing is memory management: no smart pointers are used here, unlike the rest of mfmg.

Note that the vector class must have BLAS-1 operations implemented in the class.  SimpleVector implements a CPU-memory vector; this should easily to adapt to something that stores and operates on the data on the GPU.

Let me know if you have questions.

### DIFF
--- a/include/mfmg/dealii/lanczos_deflatedlanczos.hpp
+++ b/include/mfmg/dealii/lanczos_deflatedlanczos.hpp
@@ -18,6 +18,12 @@ namespace mfmg::lanczos
 {
 
 //-----------------------------------------------------------------------------
+/// \brief Deflated Lanczos solver
+///
+///        The Lanczos solver is called multiple times.  After each Lanczos
+///        solve, the operator is modified to deflate out all previously
+///        computed (approximate) eigenvectors.  This is meant to deal with
+///        possible eigenvalue multiplicities.
 
 template<typename Op_t_>
 class DeflatedLanczos {
@@ -54,17 +60,19 @@ class DeflatedLanczos {
   private:
 
     const Op_t& op_;
-    const int num_evecs_per_cycle_;
-    const int num_cycles_;
-    const int maxit_;
-    const double tol_;
+    const int num_evecs_per_cycle_; // number of eigs to calc per lanc solve
+    const int num_cycles_;          // number of lanczos solves
+    const int maxit_;               // maximum number of lanc interations
+    const double tol_;              // convergence tolerance for eigenvalue
     const unsigned int percent_overshoot_;
-    const unsigned int verbosity_;
+                                    // allowed iteration count overshoot from
+                                    // less frequent stopping tests
+    const unsigned int verbosity_;  // verbosity of output
 
-    size_t dim_;
+    size_t dim_;                    // operator and vector dimension
 
-    std::vector<Scalar_t> evals_;
-    Vectors_t evecs_;
+    std::vector<Scalar_t> evals_;   // (approximate) eigenvals of full operator
+    Vectors_t evecs_;               // (approximate) eigenvecs of full operator
 
     // Disallowed methods
 
--- a/include/mfmg/dealii/lanczos_deflatedlanczos.template.hpp
+++ b/include/mfmg/dealii/lanczos_deflatedlanczos.template.hpp
@@ -28,6 +28,7 @@ namespace mfmg::lanczos
 {
 
 //-----------------------------------------------------------------------------
+/// \brief Deflated Lanczos solver: constructor
 
 template<typename Op_t>
 DeflatedLanczos<Op_t>::DeflatedLanczos(Op_t& op, int num_evecs_per_cycle,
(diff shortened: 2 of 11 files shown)


aprokop:
ok to test

include/mfmg/dealii/lanczos_deflatedlanczos.template.hpp
```diff
@@ -0,0 +1,151 @@
...
+ unsigned int verbosity)
+  : op_(op)
+  , num_evecs_per_cycle_(num_evecs_per_cycle)
+  , num_cycles_(num_cycles)
+  , maxit_(maxit)
+  , tol_(tol)
+  , percent_overshoot_(percent_overshoot)
+  , verbosity_(verbosity)
+  , dim_(op.dim()) {
+  assert(this->num_evecs_per_cycle_ >= 1);
```
Rombur: We have our own `MFMG_ASSERT` that we disable in release mode

include/mfmg/dealii/lanczos_deflatedlanczos.hpp
```diff
@@ -0,0 +1,89 @@
...
+  private:
+
+    const Op_t& op_;
+    const int num_evecs_per_cycle_; // number of eigs to calc per lanc solve
+    const int num_cycles_;          // number of lanczos solves
+    const int maxit_;               // maximum number of lanc interations
+    const double tol_;              // convergence tolerance for eigenvalue
+    const unsigned int percent_overshoot_;
+                                    // allowed iteration count overshoot from
+                                    // less frequent stopping tests
```
Rombur: I am not sure what this means.

include/mfmg/dealii/lanczos_deflatedlanczos.hpp
```diff
@@ -0,0 +1,89 @@
...
+    const unsigned int verbosity_;  // verbosity of output
+
+    size_t dim_;                    // operator and vector dimension
+
+    std::vector<Scalar_t> evals_;   // (approximate) eigenvals of full operator
+    Vectors_t evecs_;               // (approximate) eigenvecs of full operator
```
Rombur: Any reason why you want:
1) work with pointers
2) work with raw pointers instead of smart pointers?

include/mfmg/dealii/lanczos_lanczos.template.hpp
```diff
@@ -0,0 +1,398 @@
...
+  } // if
+
+  // Terminate
+
+  delete matrix;
+  delete t_evals_r;
+  delete t_evals_i;
+  delete t_evecs;
+
+#if 0
```
Rombur: What is this code for?

include/mfmg/dealii/lanczos_simpleop.hpp
```diff
@@ -0,0 +1,76 @@
...
+#define _LANCZOS_SIMPLEOP_HPP_
+
+#include <cstddef>
+#include <vector>
+
+namespace mfmg::lanczos
+{
+
+//-----------------------------------------------------------------------------
+/// \brief Simple test operator
```
Rombur: This should go in `test`

include/mfmg/dealii/lanczos_simpleop.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+    // Accessors
+
+    size_t dim() const {return dim_;}
+
+    Scalar_t eigenvalue(size_t i) const {return diag_value_(i);}
+
+    // Operations
+
+    void apply(Vector_t& vout, const Vector_t& vin) const;
```
Rombur: We do the apply the other way around, i.e., first `in` then `out`

include/mfmg/dealii/lanczos_deflatedlanczos.hpp
Rombur: @wdj

include/mfmg/dealii/lanczos_deflatedlanczos.hpp
Rombur: @wdj any reason why you want to use pointers?

include/mfmg/dealii/lanczos_lanczos.template.hpp
Rombur: @wdj

include/mfmg/dealii/lanczos_deflatedlanczos.hpp
wdj: No - can be converted to smart pointers as appropriate.

include/mfmg/dealii/lanczos_deflatedlanczos.hpp
wdj: A runtime-saving feature to avoid computing the eigenvalues of the tridiagonal to do the stopping test at every Lanczos step.  By computing every few steps you may save compute time but go a few more steps (no more than a fixed percentage) than needed.

include/mfmg/dealii/lanczos_lanczos.template.hpp
wdj: Code under the #if 0 - just a comment, a mental reminder of the call signature of those possibly relevant lapack routines.

PR #123, opened 2019-01-26
Apply ARPACK's patch to deal.II in CI image

dalg24:


Rombur:
Have you already pushed the new image?

dalg24:
I did.  It's already tagged as latest.

PR #124, opened 2019-01-28
[WIP]

dalg24:
@Rombur would you please look at the failures with `APPROACH3`?


Rombur:
The problem is that the eigenvectors are garbage :( I am not sure what to do about it.

Rombur:
I guess we can try shift the eigenvalues and see if that helps see https://scicomp.stackexchange.com/questions/24345/zero-eigenvalues-in-lanczos-algorithm

dalg24:
Closing in favor of #131

PR #127, opened 2019-01-31
Lanczos eigensolver

aprokop:
Supersedes #121.

Cleanup is in process


include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+#include "cblas.h"
+#include "lanczos.hpp"
+#include "lanczos_deflatedop.templates.hpp"
+
+// This complex code has to be included before lapacke for the code to compile.
+// Otherwise, it conflicts with boost or Kokkos.
+#include <complex>
+#define lapack_complex_float std::complex<float>
+#define lapack_complex_double std::complex<double>
+#include "lapacke.h"
```
dalg24: Why do you mix `#include <foo.h>` and `#include <bar.h>`?

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+lapack_int LAPACKE_dgeev( int matrix_order, char jobvl, char jobvr,
+                          lapack_int n, double* a, lapack_int lda, double* wr,
+                          double* wi, double* vl, lapack_int ldvl, double* vr,
+                          lapack_int ldvr );
+
+lapack_int LAPACKE_dgeev_work( int matrix_order, char jobvl, char jobvr,
+                               lapack_int n, double* a, lapack_int lda,
+                               double* wr, double* wi, double* vl,
+                               lapack_int ldvl, double* vr, lapack_int ldvr,
+                               double* work, lapack_int lwork );
```
dalg24: What is the deal with these declarations?  Aren't they provided by the header `<lapacke.h>`?

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+//-----------------------------------------------------------------------------
+/// \brief Lanczos solver: calculate full (approx) evecs from tridiag evecs
+template <typename OperatorType, typename VectorType>
+void Lanczos<OperatorType, VectorType>::details_calc_evecs(
+    const int num_requested, const int n,
+    std::vector<VectorType> const &lanc_vectors,
+    std::vector<ScalarType> const &evecs_tridiag,
+    std::vector<VectorType> &evecs)
+{
+  assert(evecs.size() == 0);
```
dalg24:
```suggestion
  assert(evecs.empty());
```
if you really want the assertion.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+  {
+    evecs.push_back(VectorType(dim));
+    evecs[i] = 0.0;
+
+    for (int j = 0; j < n; ++j)
+    {
+      evecs[i].add(evecs_tridiag[j + n * i], lanc_vectors[j]);
+    }
+    // evecs[i]->print();
+  }
```
dalg24: Don't bother with the `vector::push_back()`  just do
```C++
  evecs = std::vector<VectorType>(num_requested, VectorType(dim));
  for (int i = 0; i < num_requested; ++i)
  {
    for (int j = 0; j < n; ++j)
    {
      evecs[i].add(evecs_tridiag[j + n * i], lanc_vectors[j]);
    }
  }
```

tests/lanczos_simpleop.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  ScalarType diag_value_(size_t i) const
+  {
+    return (ScalarType)(1 + i / _multiplicity);
+  }
+
+  // Disallowed methods
+
+  SimpleOperator(const SimpleOperator<VectorType> &);
+  void operator=(const SimpleOperator<VectorType> &);
```
dalg24: Use `= delete;`

tests/lanczos_simpleop.templates.hpp
```diff
@@ -0,0 +1,56 @@
...
+  assert(multiplicity > 0);
+}
+
+//-----------------------------------------------------------------------------
+/// \brief Simple test operator: destructor
+
+template <typename VectorType>
+SimpleOperator<VectorType>::~SimpleOperator()
+{
+}
```
dalg24: Remove from declaration of the class.

tests/lanczos_simpleop.templates.hpp
```diff
@@ -0,0 +1,56 @@
...
+{
+
+//-----------------------------------------------------------------------------
+/// \brief Simple test operator: constructor
+
+template <typename VectorType>
+SimpleOperator<VectorType>::SimpleOperator(size_t dim, size_t multiplicity)
+    : _dim(dim), _multiplicity(multiplicity)
+{
+  // assert(this->_dim >= 0);
```
dalg24: Why is this line commented?  Do you want the assertion or not?

tests/lanczos_simpleop.templates.hpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+//-----------------------------------------------------------------------------
+/// \brief Simple test operator: apply operator to a vector
+
+template <typename VectorType>
+void SimpleOperator<VectorType>::vmult(VectorType &y, VectorType const &x) const
+{
+
+  for (int i = 0; i < this->_dim; ++i)
```
dalg24:
```suggestion
  for (int i = 0; i < _dim; ++i)
```
for consistency.  Please `grep` for `this->` and remove everywhere in this patch.

tests/lanczos_simpleop.hpp
```diff
@@ -0,0 +1,75 @@
...
+  size_t m() const { return _dim; }
+  size_t n() const { return _dim; }
+
+private:
+  size_t _dim;
+  size_t _multiplicity;
+
+  // This will be a diagonal matrix; specify the diag entries here
+
+  ScalarType diag_value_(size_t i) const
```
dalg24: I don't see the point of that member function.  Why don't you just move the definition into `eigenvalue()`.

tests/test_lanczos.cc
```diff
@@ -0,0 +1,125 @@
...
+
+BOOST_AUTO_TEST_CASE(lanczos)
+{
+  using namespace mfmg;
+
+  typedef double ScalarType;
+  typedef dealii::Vector<ScalarType> VectorType;
+  typedef SimpleOperator<VectorType> OperatorType;
+
+  typedef Lanczos<OperatorType, VectorType> Solver_t;
```
dalg24: Prefer C++11 type aliasing to `typedef`.  Also `Solver_t` does not follow coding style in this project.

tests/test_lanczos.cc
```diff
@@ -0,0 +1,125 @@
...
+  // std::cout << std::endl;
+
+  boost::property_tree::ptree lanczos_params;
+  lanczos_params.put("is_deflated", true);
+  lanczos_params.put("num_eigenpairs_per_cycle", num_evecs_per_cycle);
+  lanczos_params.put("num_cycles", num_cycles);
+  lanczos_params.put("max_iterations", maxit);
+  lanczos_params.put("tolerance", tol);
+  lanczos_params.put("percent_overshoot", percent_overshoot);
+  lanczos_params.put("verbosity", verbosity);
```
dalg24: That's silly.  Do not declare variables to put them into the property tree later.  Just put them directly in place.

tests/test_lanczos.cc
```diff
@@ -0,0 +1,125 @@
...
+
+  unsigned int rank = dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
+  BOOST_TEST(rank == 0); // only need to test on 1 rank
+
+  // Loop to ensure each Ritz value is near an eigenvalue.
+  // TODO: make a more complete check of correctness of result.
+
+  const int num_ritzvals = solver.num_evecs();
+  int num_passed = 0;
+  const double tol_test = 1. * tol; // this may need adjustment
```
dalg24: ???

tests/test_lanczos.cc
```diff
@@ -0,0 +1,125 @@
...
+    std::cout.width(4);
+    std::cout.precision(8);
+    std::cout << std::fixed << solver.get_eval(i) << " ";
+  }
+  std::cout << std::endl;
+
+  //-----
+
+  unsigned int rank = dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
+  BOOST_TEST(rank == 0); // only need to test on 1 rank
```
dalg24: Don't use `BOOST_TEST()` for that!

tests/test_lanczos.cc
```diff
@@ -0,0 +1,125 @@
...
+  // TODO: make a more complete check of correctness of result.
+
+  const int num_ritzvals = solver.num_evecs();
+  int num_passed = 0;
+  const double tol_test = 1. * tol; // this may need adjustment
+
+  for (int i = 0; i < num_ritzvals; ++i)
+  {
+    double ritzval = solver.get_eval(i);
+    for (int j = 0; j < (int)n; ++j)
```
dalg24: Boo!  Why did you declare `n` unsigned in the first place?

tests/test_lanczos.cc
```diff
@@ -0,0 +1,125 @@
...
+#include "main.cc"
+
+//-----------------------------------------------------------------------------
+
+BOOST_AUTO_TEST_CASE(lanczos)
+{
+  using namespace mfmg;
+
+  typedef double ScalarType;
+  typedef dealii::Vector<ScalarType> VectorType;
```
dalg24:
```suggestion
  typedef dealii::Vector<double> VectorType;
```
Don't bother with `ScalarType`.

tests/test_lanczos.cc
```diff
@@ -0,0 +1,125 @@
...
+    {
+      const double eval = op.eigenvalue(j);
+      const double diff = fabs(ritzval - eval);
+      const bool match = diff < tol_test;
+      // std::cout << i << " " << eval << " " << ritzval << " " << "\n";
+      if (match)
+      {
+        num_passed++;
+        break;
+      }
```
dalg24: I was about to make compliments about the linear algebra in this PR.  I take it all back!

What is happening here???
First of drop that `num_passed++` and use the `BOOST_TEST` assertion macros directly in the loop body.
Second , what's the deal here, how are all `{op.eigenvalue(j) for j in [0, n)}` equal to `solver.get_eval(i)`?
Is `solver.get_eval(i)` a constant?  Cause `SimpleOperator::eigenvalue(j)` should vary with `j`...

I just don't get it.

dalg24: Nevermind I think I understand now.

Assuming that we add `std::vector<ScalarType> const &Lanczos::get_evals() const;` and that we somehow get a (unique) list of the eigenvalues from `SimpleVector`, say `ref_evals`.
```C++
auto ritzvals = solver.get_evals();
namespace fpc = boost::math::fpc;
auto close_at_tolerance = fpc::close_at_tolerance<double>(fpc::percent_tolerance(0.1));
for (auto eval : ref_evals)
{
  TEST_ASSERT(std::count_if(
      ritzvals.begin(), ritzvals.end(),
        [&](double v) { return close_at_tolerance(v, eval); }) == multiplicity);
}
```

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -185,6 +186,35 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
...
+    lanczos_params.put("tolerance", tolerance);
+    lanczos_params.put("percent_overshoot", 5);
+    lanczos_params.put("verbosity", 0);
+
+    Lanczos<dealii::SparseMatrix<double>, dealii::Vector<double>> solver(
+        agglomerate_system_matrix, lanczos_params);
+
+    solver.solve();
+    ASSERT(n_eigenvectors == solver.num_evecs(),
+           "Wrong number of computed eigenpairs");
```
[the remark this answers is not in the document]
aprokop: This is a good question that I don't know how to answer. It was also the reason why I switched to `#if 0`. I need to further understand how to get the exact number of required eigenvectors when running deflated Lanczos.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+
+  if (!_is_deflated)
+  {
+    // By default set initial guess to a random vector.
+    VectorType initial_guess(_op.n());
+    {
+      std::mt19937 gen(0);
+      std::uniform_real_distribution<double> dis(0, 1);
+      for (int i = 0; i < n; i++)
+        initial_guess[i] = (ScalarType)(dis(gen));
```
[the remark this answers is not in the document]
aprokop: Many libraries do not provide such access.

tests/lanczos_simpleop.hpp
aprokop: Because diagonal value is also used in `apply`.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+  // but instead recalculate them for this use.
+  // However this may be dangerous if the second lanczos iteration
+  // has different roundoff characteristics, e.g., due to order of
+  // operatioon differences.
+  // ISSUE: we have not taken precautions here with regard to
+  // potential impacts of loss of orthogonality of lanczos vectors.
+  details_calc_evecs(num_requested, it, lanc_vectors, evecs_tridiag, evecs);
+
+  if (_verbosity > 0)
+    std::cout << std::endl;
```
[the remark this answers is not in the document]
aprokop: What do you have against `endl`?

include/mfmg/dealii/lanczos.templates.hpp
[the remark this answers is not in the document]
dalg24: > Many libraries do not provide such access.

I don't think this is true.  Also a pointers would do if the elements are stored in a contiguous fashion.

include/mfmg/dealii/lanczos.templates.hpp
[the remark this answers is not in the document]
dalg24: What you want is `'\n'` not the `std::cout::flush()`.  Also this "if verbose mode is on, insert a new line" looks silly.

tests/lanczos_simpleop.hpp
dalg24: So?  `SimpleOperator::vmult()` could call directly `eigenvalue()`.   I just didn't see the point of having two member functions but I suppose it's a way of declaring an alias of the function and I can see that either names is more suitable depending on the context.  I guess I can live with it :)

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+                     "---------------------------------------"
+                  << std::endl;
+        std::cout << "Lanczos solve " << cycle + 1 << ":" << std::endl;
+      }
+
+      // Perform Lanczos solve, initial guess is a lin comb of a
+      // constant vector (to try to capture "smooth" eigenmodes
+      // of PDE problems and a random vector based on different random
+      // seeds.
+      // ISSUE; should a different initial guess strategy be used.
```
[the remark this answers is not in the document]
aprokop: It is up date. It is talking about possibly different way to initialize initial guess.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+      VectorType initial_guess(_op.n());
+      {
+        std::mt19937 gen(cycle);
+        std::uniform_real_distribution<double> dis(0, 1);
+        for (int i = 0; i < initial_guess.size(); i++)
+          initial_guess[i] = (ScalarType)(1. + dis(gen));
+      }
```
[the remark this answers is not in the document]
aprokop: Not sure what you mean.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+
+template <typename OperatorType, typename VectorType>
+template <typename FullOperatorType>
+void Lanczos<OperatorType, VectorType>::details_solve_lanczos(
+    FullOperatorType const &op, const int num_requested,
+    VectorType const &initial_guess, std::vector<ScalarType> &evals,
+    std::vector<VectorType> &evecs)
+{
+  // FIXME: right now Operator does not have dimension accessors
+  const int n = _op.m();
```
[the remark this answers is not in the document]
aprokop: Yes, I did not go all the way to make them truly independent. I only wanted to simplify the flow of the real data around to ease readability.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+
+  std::vector<ScalarType>
+      evecs_tridiag; // eigenvecs of tridiag matrix, stored in flat array
+
+  // Lanczos iteration loop.
+  int it = 1;
+  for (int it_prev_check = 0; it <= _maxit; ++it)
+  {
+    // Normalize lanczos vector.
+    assert(beta != 0); // TODO: set up better check for near-zero.
```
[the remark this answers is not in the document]
aprokop: I don't know how to choose this tolerance.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,494 @@
...
+  assert(t_offdiag.size() == n - 1);
+
+  // Allocate storage
+  std::vector<double> matrix(n * n, 0.);
+  std::vector<double> t_evals_r(n, 0.);
+  std::vector<double> t_evals_i(n, 0.);
+  std::vector<double> t_evecs(n * n, 0.);
+
+  // Copy diagonals of the tridiagonal matrix into the full matrix
+  matrix[0] = (double)t_maindiag[0];
```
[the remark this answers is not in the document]
aprokop: I would be fine removing all `ScalarType` references, but need an OK from @wdj.

wdj:
is ok --

aprokop:
@dalg24 Addressed most of your comments.

aprokop:
OK. The code can be considered cleaned up. Now need to work on the test. Currently it fails for the deflation mode: instead of finding 1, 1, 2, 2, 3, 3, it finds 1, 1, 2, 2, 3, 4.

include/mfmg/dealii/lanczos.templates.hpp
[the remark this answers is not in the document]
dalg24: Keep it but add a static assertion that it is the same as `double`.  And then let's get rid of the mess with all these explicit conversions, that mixes C-style cast and `static_cast`...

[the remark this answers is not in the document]
aprokop: Too late.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -166,9 +141,7 @@ void Lanczos<OperatorType, VectorType>::solve()
 
       // Add eigenvectors to the set of vectors being deflated out
       if (cycle != _num_cycles - 1)
-      {
         deflated_op.add_deflation_vecs(evecs);
-      }
```
dalg24: Leave the brackets.  clang-tidy emits a readability warning without it.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -363,44 +316,25 @@ void Lanczos<OperatorType, VectorType>::details_calc_tridiag_epairs(
...
-
   // Terminate if every approximage eigenvalue has converged to tolerance
   // ISSUE: here ignoring possible nuances regarding the correctness
   // of this check.
   // NOTE: k may be desirable to "scale" the stopping criterion
   // based on (estimate of) matrix norm or similar.
   for (int i = 0; i < num_requested; ++i)
   {
-    const double bound = beta * abs(evecs[n - 1 + n * i]);
+    const double bound = beta * fabs(evecs[n - 1 + n * i]);
```
dalg24: They are `std::abs` overloads for floating-pointer numbers.  Why did you change?

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -430,6 +365,20 @@ void Lanczos<OperatorType, VectorType>::details_calc_evecs(
...
+// and a random vector based on different random seeds.
+// ISSUE: should a different initial guess strategy be used?
+template <typename OperatorType, typename VectorType>
+void Lanczos<OperatorType, VectorType>::details_set_initial_guess(
+    VectorType &initial_guess, int seed) const
+{
+  std::mt19937 gen(seed);
+  std::uniform_real_distribution<double> dist(0, 1);
+  std::generate(initial_guess.begin(), initial_guess.end(),
+                [&]() { return 1 + dist(gen); });
```
dalg24: I didn't catch that before.  Just do `std::uniform_real_distribution<double> dist(1, 2);` and `{ return dist(gen); }`

include/mfmg/dealii/lanczos.templates.hpp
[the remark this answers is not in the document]
dalg24: `git reset HEAD~1`

include/mfmg/dealii/lanczos.templates.hpp
aprokop: No, this way the intention is more clear.

dalg24: You can't be serious.  Saying you want a random real number between 0 and 1 and then shifting it by one is more clear than generating it between 1 and 2??

dalg24: 1 slwikdf7w3-unnnuuxkulw

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -185,6 +186,34 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
     solver.solve(agglomerate_system_matrix, agglomerate_mass_matrix,
                  inv_system_matrix, eigenvalues, eigenvectors);
   }
+  else if (_eigensolver_type == "lanczos")
+  {
+    boost::property_tree::ptree lanczos_params;
+#if 0
```
dalg24: If you really are going to keep these preprocessor directives, define something like LANCZOS_IS_DEFLATED and use it here.

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -185,6 +186,34 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
...
+    Lanczos<dealii::SparseMatrix<double>, dealii::Vector<double>> solver(
+        agglomerate_system_matrix, lanczos_params);
+
+    solver.solve();
+    ASSERT(n_eigenvectors == static_cast<unsigned int>(solver.num_evecs()),
+           "Wrong number of computed eigenpairs");
+
+    // Copy the eigenvalues and the eigenvectors
+    for (unsigned int i = 0; i < n_eigenvectors; ++i)
+      eigenvalues[i] = solver.get_eval(i);
```
dalg24: Replace loop by `eigenvalues = solver.get_evals();`

include/mfmg/dealii/lanczos.hpp
```diff
@@ -0,0 +1,98 @@
...
+//-----------------------------------------------------------------------------
+/// \brief Lanczos solver
+
+template <typename OperatorType, typename VectorType>
+class Lanczos
+{
+public:
+  // Ctor/dtor
+  Lanczos(OperatorType const &op, boost::property_tree::ptree const &params);
+  ~Lanczos() {}
```
dalg24:
```suggestion
```

include/mfmg/dealii/lanczos.hpp
```diff
@@ -0,0 +1,98 @@
...
+  void details_calc_evecs(const int num_requested, const int n,
+                          std::vector<VectorType> const &lanc_vectors,
+                          std::vector<double> const &evecs_tridiag,
+                          std::vector<VectorType> &evecs) const;
+
+  void details_set_initial_guess(VectorType &v, int seed = 0) const;
+
+  // Disallowed methods
+
+  Lanczos(const Lanczos<OperatorType, VectorType> &);
```
dalg24: Lanczos(const Lanczos<OperatorType, VectorType> &) = delete;

include/mfmg/dealii/lanczos.hpp
```diff
@@ -0,0 +1,98 @@
...
+  void operator=(const Lanczos<OperatorType, VectorType> &);
```
dalg24: What is that operator?

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,384 @@
...
+    _num_requested = params.get<int>("num_eigenpairs");
+
+    assert(_num_requested >= 1);
+    assert(_maxit >= _num_requested &&
+           "maxit too small to produce required number of eigenvectors.");
+  }
+
+  assert(_op.m() == _op.n() && "Operator must be square");
+
+  assert(_maxit >= 0);
```
dalg24: This assertion seems superfluous since you have an assert for `_maxit` in both branches above.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+namespace mfmg
+{
+
+/// \brief Lanczos solver: constructor
+template <typename OperatorType, typename VectorType>
+Lanczos<OperatorType, VectorType>::Lanczos(
+    OperatorType const &op, boost::property_tree::ptree const &params)
+    : _op(op)
+{
```
dalg24: I am not a big fan of the data member initialization in that constructor.
Some of the data members are uninitialized depending on the value of `_is_deflated` but it is not obvious when looking at the body.
I would suggest `: _op{op}, _is_deflated{params.get<bool>("is_deflated")}, ..., _num_evecs_per_cycle{_is_deflated ?params.get<int>("num_eigenpairs_per_cycle") : <some value possibly denoting it is invalid>}, ... { <constructor body with assertions, preferably no if> }`

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,384 @@
...
+  // library.
+  // NOTE: for accuracy we are using double here rather than double;
+  // may not be needed.
+  // ISSUE: LAPACK has other solvers that might be more efficient here.
+
+  // Do all in double, regardless of double (for now)
+  const lapack_int info = LAPACKE_dgeev(
+      LAPACK_COL_MAJOR, 'N', 'V', n, matrix.data(), n, t_evals_r.data(),
+      t_evals_i.data(), NULL, n, t_evecs.data(), n);
+  ASSERT(!info, "Call to LAPACKE_dgeev failed.");
```
dalg24: Returns non-zero on success :(

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+    // Check convergence if requested
+    // NOTE: an alternative here for p > 0 is
+    // int((100./p)*ln(it)) > int((100./p)*ln(it-1))
+    const bool do_check =
+        it == 1 || it == _maxit ||
+        (100 * (it - it_prev_check) > _percent_overshoot * it_prev_check);
+    if (do_check)
+    {
+      // Calculate eigenpairs of tridiagonal matrix for convvergence test or at
```
dalg24: s/vv/v/

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,384 @@
...
+  details_calc_evecs(num_requested, it, lanc_vectors, evecs_tridiag, evecs);
+}
+
+/// \brief Lanczos solver: calculate eigenpairs from tridiagonal of Lanczos
+/// coefficients
+template <typename OperatorType, typename VectorType>
+void Lanczos<OperatorType, VectorType>::details_calc_tridiag_epairs(
+    std::vector<double> const &t_maindiag, std::vector<double> const &t_offdiag,
+    const int num_requested, std::vector<double> &evals,
+    std::vector<double> &evecs) const
```
dalg24: Looks like this one can be static too.

Readability: `evals` and `evecs` are out parameters so return them instead of passing them by reference.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,384 @@
...
+  const lapack_int info = LAPACKE_dgeev(
+      LAPACK_COL_MAJOR, 'N', 'V', n, matrix.data(), n, t_evals_r.data(),
+      t_evals_i.data(), NULL, n, t_evecs.data(), n);
+  ASSERT(!info, "Call to LAPACKE_dgeev failed.");
+
+  // Compute permutation for ascending order
+  std::vector<int> perm_index(n);
+  std::iota(perm_index.begin(), perm_index.end(), 0);
+  std::sort(perm_index.begin(), perm_index.end(),
+            [&](int i, int j) { return t_evals_r[i] < t_evals_r[j]; });
```
dalg24: Don't bother computing the permutation order if `n < num_requested`

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,384 @@
...
+  }
+}
+
+/// \brief Lanczos solver: perform Lanczos solve
+template <typename OperatorType, typename VectorType>
+template <typename FullOperatorType>
+void Lanczos<OperatorType, VectorType>::details_solve_lanczos(
+    FullOperatorType const &op, const int num_requested,
+    VectorType const &initial_guess, std::vector<double> &evals,
+    std::vector<VectorType> &evecs) const
```
dalg24: Readability: consider returning `evals` and `evecs` instead of taking them by reference.

include/mfmg/dealii/lanczos_deflatedop.templates.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include "lanczos_deflatedop.hpp"
+
+namespace mfmg
+{
+
+/// \brief Deflated operator: constructor
+template <typename OperatorType, typename VectorType>
+DeflatedOperator<OperatorType, VectorType>::DeflatedOperator(
+    const OperatorType &base_op)
```
dalg24: Comment that you are using curiously recurring template pattern here.

include/mfmg/dealii/lanczos_deflatedop.hpp
```diff
@@ -0,0 +1,59 @@
...
+  void deflate(VectorType &vec) const;
+
+private:
+  OperatorType const &_base_op; // reference to the base operator object
+
+  std::vector<VectorType> _deflation_vecs; // vectors to deflate out
+
+  // Disallowed methods
+  DeflatedOperator(const DeflatedOperator<OperatorType, VectorType> &) = delete;
+  void operator=(const DeflatedOperator<OperatorType, VectorType> &) = delete;
```
dalg24: What is that operator?

Move the declaration of the deleted copy constructor next to the other constructor.

include/mfmg/dealii/lanczos_deflatedop.templates.hpp
dalg24: Would you please recall on this thread what are the requirements for `OperatorType`?
`const`-qualified `m()`, `n()`, and `vmult()` member functions.  Is that all?

include/mfmg/dealii/lanczos.templates.hpp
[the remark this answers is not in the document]
aprokop: Not doing this.

include/mfmg/dealii/lanczos_deflatedop.templates.hpp
aprokop: yes.

include/mfmg/dealii/amge_host.templates.hpp
aprokop: No, the have different type (`complex` vs `double`).

aprokop:
I think this is ready for the final review. It still has issues, but I think they are numerical analysis issues rather than code.

I think I also need to add more testing for this PR to test for integration.

include/mfmg/dealii/lanczos.hpp
```diff
@@ -0,0 +1,70 @@
...
+ * 3-clause license. For the licensing terms see the LICENSE file in the *
+ * top-level directory                                                   *
+ *                                                                       *
+ * SPDX-License-Identifier: BSD-3-Clause                                 *
+ *************************************************************************/
+
+#ifndef MFMG_LANCZOS_LANCZOS_HPP
+#define MFMG_LANCZOS_LANCZOS_HPP
+
+#include <mfmg/common/operator.hpp>
```
dalg24: Header not needed.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,342 @@
...
+ *                                                                       *
+ * SPDX-License-Identifier: BSD-3-Clause                                 *
+ *************************************************************************/
+
+#ifndef MFMG_LANCZOS_LANCZOS_TEMPLATE_HPP
+#define MFMG_LANCZOS_LANCZOS_TEMPLATE_HPP
+
+#include <algorithm>
+#include <cassert>
+#include <iostream>
```
dalg24: Not needed. You don't use IO any more since you removed the verbose mode.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,342 @@
...
+ * top-level directory                                                   *
+ *                                                                       *
+ * SPDX-License-Identifier: BSD-3-Clause                                 *
+ *************************************************************************/
+
+#ifndef MFMG_LANCZOS_LANCZOS_TEMPLATE_HPP
+#define MFMG_LANCZOS_LANCZOS_TEMPLATE_HPP
+
+#include <algorithm>
+#include <cassert>
```
dalg24: Probably should move mfmg's assertion "macros".

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,342 @@
...
+    // Check convergence if requested
+    // NOTE: an alternative here for p > 0 is
+    // int((100./p)*ln(it)) > int((100./p)*ln(it-1))
+    bool const first_iteration = (it == 1);
+    bool const max_iterations_reached = (it == maxit);
+    bool const percent_overshoot_exceeded =
+        (100 * (it - it_prev_check) > percent_overshoot * it_prev_check);
+    if (first_iteration || max_iterations_reached || percent_overshoot_exceeded)
+    {
+      const int dim_hessenberg = it;
```
dalg24: Declare it unsigned and the hell with cast in the assertion below.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -0,0 +1,342 @@
...
+{
+  auto dim = lanc_vectors[0].size();
+
+  std::vector<VectorType> evecs(num_requested, VectorType(dim));
+
+  // Matrix-matrix product to convert tridiagonal eigenvectors to operator
+  // eigenvectors
+  for (int i = 0; i < num_requested; ++i)
+  {
+    evecs[i] = 0.0;
```
dalg24: Unnecessary since the vector is already initialized to zero.

include/mfmg/dealii/lanczos_deflatedop.templates.hpp
```diff
@@ -0,0 +1,130 @@
...
+        dot_best = dot_this;
+        std::swap(perm_ind[i], perm_ind[j]);
+      }
+    }
+
+    // Normalize.
+
+    // FIXME: we are not accounting for possible rank deficiency here
+    double norm = std::sqrt(dot_best);
+    assert(norm); // FIXME need better test for near-zero here.
```
dalg24: Readability: `assert(norm == 0.);`

tests/lanczos_simpleop.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * top-level directory                                                   *
+ *                                                                       *
+ * SPDX-License-Identifier: BSD-3-Clause                                 *
+ *************************************************************************/
+
+#ifndef MFMG_LANCZOS_SIMPLEOP_HPP
+#define MFMG_LANCZOS_SIMPLEOP_HPP
+
+#include <mfmg/common/exceptions.hpp>
+#include <mfmg/common/operator.hpp>
```
dalg24: Doesn't derive from `mfmg::Operator` does it?

tests/lanczos_simpleop.templates.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *                                                                       *
+ * SPDX-License-Identifier: BSD-3-Clause                                 *
+ *************************************************************************/
+
+#ifndef MFMG_LANCZOS_SIMPLEOP_TEMPLATE_HPP
+#define MFMG_LANCZOS_SIMPLEOP_TEMPLATE_HPP
+
+#include <cassert>
+
+#include "lanczos_simpleop.hpp"
```
dalg24: Consistency: Use angle brackets.

tests/lanczos_simpleop.hpp
```diff
@@ -0,0 +1,61 @@
...
+private:
+  size_t _dim;
+  size_t _multiplicity;
+
+  // This will be a diagonal matrix; specify the diag entries here
+  double diag_value_(size_t i) const { return 1 + i / _multiplicity; }
+
+  // Disallowed methods
+  SimpleOperator(const SimpleOperator<VectorType> &) = delete;
+  void operator=(const SimpleOperator<VectorType> &) = delete;
```
dalg24: :(

tests/test_lanczos.cc
```diff
@@ -0,0 +1,83 @@
...
+#include "lanczos_simpleop.templates.hpp"
+#include "main.cc"
+
+namespace bdata = boost::unit_test::data;
+namespace tt = boost::test_tools;
+
+BOOST_DATA_TEST_CASE(lanczos,
+                     bdata::make({1, 2}) * bdata::make({false, true}) *
+                         bdata::make({1, 2, 3, 5}),
+                     multiplicity, is_deflated, n_distinct_eigenvalues)
```
dalg24: Good!

tests/test_lanczos.cc
```diff
@@ -0,0 +1,83 @@
...
+  std::vector<double> computed_evals;
+  std::vector<VectorType> computed_evecs;
+  std::tie(computed_evals, computed_evecs) = solver.solve(lanczos_params);
+
+  auto ref_evals = op.get_evals();
+
+  BOOST_TEST(computed_evals.size() == n_eigenvectors);
+
+  // Loop to ensure each Ritz value is near an eigenvalue.
+  // TODO: make a more complete check of correctness of result
```
dalg24: By that you mean look at the eigenvectors?

include/mfmg/dealii/lanczos.templates.hpp
aprokop: It would fix the assert, but would cascade later for comparisons with `num_requested` and such. Just leaving this alone.

include/mfmg/dealii/lanczos.templates.hpp
aprokop: I was not sure if dealii does that. I would also like to leave it alone to make things clear.

include/mfmg/dealii/lanczos_deflatedop.templates.hpp
aprokop: Seriously? Did you mean `assert(norm != 0.)`?

tests/lanczos_simpleop.templates.hpp
aprokop: Hmm, I tried that, but it does not find the header if I switch to angle brackets...

wdj:
Yes, it's really just a placeholder for putting in a valid near-zero test --

tests/lanczos_simpleop.templates.hpp
Rombur: Yeah, the headers that are only in `tests` are not exposed correctly. This is the correct way to do it.

wdj:
A more valid test would use machine epsilon, ~ 1e-15, and a bound on the norm of A --

include/mfmg/dealii/lanczos_deflatedop.templates.hpp
dalg24: Yup :)

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -102,6 +102,29 @@ Lanczos<OperatorType, VectorType>::solve(
...
+    std::sort(perm_index.begin(), perm_index.end(),
+              [&](int i, int j) { return evals[i] < evals[j]; });
+
+    for (int i = 0; i < n_eigenvectors; i++)
+    {
+      sorted_evals[i] = evals[perm_index[i]];
+      sorted_evecs[i] = evecs[perm_index[i]];
+    }
+    std::swap(evals, sorted_evals);
+    std::swap(evecs, sorted_evecs);
```
dalg24: Readability: just assign or if you care about performance do `var = std::move(sorted_var);`

tests/test_lanczos.cc
```diff
@@ -79,4 +79,12 @@ BOOST_DATA_TEST_CASE(lanczos,
...
   for (int i = 0; i < n_eigenvectors; i++)
     BOOST_TEST(computed_evals[i] == ref_evals[i], tt::tolerance(tolerance));
+
+  for (int i = 0; i < n_eigenvectors; i++)
+  {
+    VectorType result(n);
+    op.vmult(result, computed_evecs[i]);
+    result.add(-computed_evals[i], computed_evecs[i]);
+    BOOST_TEST(result.l2_norm() < tolerance);
+  }
```
dalg24: Excellent!

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -233,6 +241,10 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
       for (unsigned int j = 0; j < n_dofs_agglomerate; ++j)
         eigenvectors[i][j] = lapack_eigenvectors[j][i];
   }
+  else
+  {
+    ASSERT(true, "Unknown eigensolver type");
+  }
```
dalg24: @Rombur was there a rational for not having a `else`-clause?

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -128,7 +128,8 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
   std::vector<dealii::Vector<double>> eigenvectors(
       n_eigenvectors, dealii::Vector<double>(n_dofs_agglomerate));
 
-  if (_eigensolver_type == "arpack")
+  auto eigensolver_type = _eigensolver_params.get("type", "arpack");
```
dalg24: `const` it.  Also I would feel more comfortable if you specified `<std::string>`.  I am not sure what the deduction rule is.  I sure hope you don't get a `char *`...

include/mfmg/dealii/amge_host.templates.hpp
Rombur: Not really.

dalg24: Doesn't mean you shouldn't do that.  Just putting it here for the record.

include/mfmg/dealii/amge_host.templates.hpp
dalg24: (By that I mean `ptree::get<std::string>(...)`)

dalg24: Quick try in Cling says the return type is `boost::enable_if<boost::property_tree::detail::is_character<char>, std::basic_string<char, std::char_traits<char>, std::allocator<char> > >::type` so you're good :/

dalg24:
Good job @aprokop and @wdj

PR #128, opened 2019-02-04
Install OpenBLAS manually because the package manager does not install lapacke

Rombur:
Wayne's solver uses lapacke but it was not installed by the package manager, so we install by hand.


Rombur:
I couldn't make the manual install to work which prompted me to search if there was another way to get `lapacke` and there actually is a `lapacke-dev` package. It doesn't `OpenBLAS` but it works :)

PR #130, opened 2019-02-12
Install OpenBLAS manually

Rombur:


Rombur:
This seems to be working. The trick was to use `DYNAMIC_ARCH=1`.

aprokop:
I'm going to test the container with Lanczos branch now.

ci/Dockerfile
```diff
@@ -235,6 +262,10 @@ RUN export DEAL_II_HASH=9cd42e465767c5a8ceffddbbc17f6737503337ac && \
     rm -rf ${DEAL_II_SOURCE_DIR}
 ENV DEAL_II_DIR=${INSTALL_DIR}/dealii
 
+# deal.II does not export the include directory for BLAS/LAPACK, so add it to
+# CPATH
+ENV CPATH=${OPENBLAS_DIR}/include:${CPATH}
```
dalg24: :/

aprokop: Yeah, that. May be a bug in DealII.

Issue #114, opened 2019-01-10
Add matrix-free tests


Rombur:

dalg24:
1. make eigensolvers work for matrix free (blocked by #131)
2. implement matrix-free version of the mesh evaluator that does not derive from the regular one
3. provide a new version for all mesh evaluators in the unit tests :(

dalg24:
> 1. make eigensolvers work for matrix free (blocked by #115)
> 2. implement matrix-free version of the mesh evaluator that does not derive from the regular one
> 3. provide a new version for all mesh evaluators in the unit tests :( (blocked by 2.)

First point has been addressed in #127 and #136.  Second one is partially covered by #137.

PR #131, opened 2019-02-16
Enable matrix-free mode for eigensolver

dalg24:
related to #114

### DIFF
--- a/include/mfmg/dealii/amge_host.templates.hpp
+++ b/include/mfmg/dealii/amge_host.templates.hpp
@@ -26,6 +26,8 @@
 
 #include <EpetraExt_MatrixMatrix.h>
 
+#define MATRIX_FREE 1
+
 // APPROACH1: use UMFPACK to invert the matrix
 // APPROACH2: use CG to invert
 // APPROACH3: do not need invert, use ARPACK regular mode instead of
@@ -222,6 +224,10 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
   }
   else if (eigensolver_type == "lapack")
   {
+#if MATRIX_FREE
+    throw std::runtime_error(
+        "LAPACK not available as eigensolver in matrix-free mode");
+#else
     // Use Lapack to compute the eigenvalues
     dealii::LAPACKFullMatrix<double> full_matrix;
     full_matrix.copy_from(agglomerate_system_matrix);
@@ -241,6 +247,7 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
     for (unsigned int i = 0; i < n_eigenvectors; ++i)
       for (unsigned int j = 0; j < n_dofs_agglomerate; ++j)
         eigenvectors[i][j] = lapack_eigenvectors[j][i];
+#endif
   }
   else
   {
--- a/include/mfmg/dealii/amge_host.templates.hpp
+++ b/include/mfmg/dealii/amge_host.templates.hpp
@@ -28,16 +28,10 @@
 
 #define MATRIX_FREE 1
 
-// APPROACH1: use UMFPACK to invert the matrix
-// APPROACH3: do not need invert, use ARPACK regular mode instead of
-//            shift-and-invert mode
-
-#define APPROACH1
-
 namespace mfmg
 {
 
-#if defined(APPROACH3)
+#if MATRIX_FREE
 struct NoOp
 {
   template <typename VectorType>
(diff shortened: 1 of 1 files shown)


dalg24:
matrix-free mode:
* `test_hierarchy_1` failing because it attempts to use LAPACK as eigensolver
* `test_restriction_matrix_{1,2,3}` is more serious trouble (see [discussion on #124](https://github.com/ORNL-CEES/mfmg/pull/124#issuecomment-458587028))

dalg24:
Turning matrix-free mode off, intending like to merge "as is".

Next steps:
1. decide whether to drop matrix-free mode for Arpack -- @Rombur can you confirm
2. continuation of #127 integrate Lanczos solver in our existing tests -- assign to @aprokop?
3. make matrix-free mode work with Lancsoz (and Arpack depending on 1.) -- assign to @dalg24 or @aprokop 
  comment: blocked by 2. 
4. provide `MatrixMeshEvaluator` that does not derive from the regular `MeshEvaluator` -- assign to @dalg24?
  comment: orthogonal 2. and 3.

dalg24:
Need review

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -76,10 +48,48 @@ struct NoOp
...
+    }
+    return diag_elements;
+  }
+
+  SizeType m() const { return _matrix.m(); }
+
+  SizeType n() const { return _matrix.n(); }
+
+private:
+  dealii::SparsityPattern sparsity_pattern;
```
aprokop: `_sparsity_pattern`.

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -76,10 +48,48 @@ struct NoOp
...
+{
+  using MatrixType = dealii::SparseMatrix<double>;
+  using SizeType = MatrixType::size_type;
+
+  template <typename MeshEvaluator, typename DoFHandler>
+  NotAMatrix(MeshEvaluator const &mesh_evaluator, DoFHandler &dof_handler,
+             dealii::AffineConstraints<double> &constraints)
+  {
+    mesh_evaluator.evaluate_agglomerate(dof_handler, constraints,
+                                        sparsity_pattern, _matrix);
```
aprokop: Assert `_matrix.m() == _matrix.n()`?

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -76,10 +48,48 @@ struct NoOp
     std::ignore = src;
     std::ignore = dst;
     // Raise an error to make sure nobody uses this by inadvertance
-    throw std::runtime_error("should never get here");
+    throw std::logic_error("should never get here");
   }
 };
-#endif
+
+struct NotAMatrix
```
aprokop: `NotAMatrix` means nothing to me. Any more informative name?

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -26,48 +26,20 @@
...
-#define APPROACH1
+#define MATRIX_FREE 0
 
 namespace mfmg
 {
 
-#if defined(APPROACH2)
-template <typename MatrixType>
-struct WrapInverse
+struct Identity
```
aprokop: I would add a constructor with size to store. So it's going to be an identity of specific size. And then in `vmult` I would check that the sizes of the vectors match to this size.

dalg24: Won't fix.  I want to use the default constructor.

include/mfmg/dealii/amge_host.templates.hpp
dalg24: It is not a matrix.  It demonstrates it works without a matrix.  What other name would you like better?

include/mfmg/dealii/amge_host.templates.hpp
dalg24: Will fix.

include/mfmg/dealii/amge_host.templates.hpp
dalg24: Nah.  Will be caught by the eigensolver if it requires it.

aprokop: But it may segfault somewhere when you get diagonal values and your matrix is tall and skinny (though I do not envision this use case) as you use `m()` when getting diagonal values.

include/mfmg/dealii/amge_host.templates.hpp
aprokop: This `NotAMatrix` does not work without a matrix, and in fact stores the matrix explicitly inside itself. Maybe `MatrixBasedOperator`, or `MatrixWrapOperator`?

dalg24:
I renamed `NotAMatrix` and added the assertion that we have a square matrix.

Rombur:
> 1. decide whether to drop matrix-free mode for Arpack -- assign to @Rombur

Yes, we will need to drop it. The new eigensolver does work better so we will focus on that one only.

aprokop:
Do you want to clean the history, or can it be merged in its current form?

PR #133, opened 2019-02-19
Make sure the diffusion is always positive

Rombur:
For some geometry (like `hyper_ball`) it was possible that the diffusion coefficient would change sign -> the system would be not SPD anymore


aprokop:
Do you need to update `DiscontinuousMaterialProperty`?

Rombur:
@aprokop No the value is always 10 or 100

aprokop:
@Rombur Would `floor(p[d]*100)` with `p[d]` negative cause issues?

Rombur:
We cast it to an `unsigned int` so I don't see how it's a problem. I can change it if you want.

PR #134, opened 2019-02-20
Add MatrixMarket output for dealii::SparseMatrix

aprokop:


dalg24:
Changes proposed are reasonable.  Question though:  Where do you use/plan on using this?

aprokop:
@dalg24 At one point I used it to print out the agglomerate matrices to check their eigenvalues in Octave.

dalg24:
> @dalg24 At one point I used it to print out the agglomerate matrices to check their eigenvalues in Octave.

I don't have a strong opinion on this but I am usually not so keen on merging code that is not used.  Go ahead and merge if you feel you really want that patch.

PR #138, opened 2019-03-12
Update deal.II version in image

Rombur:


dalg24:
Is it just a WIP?

Rombur:
Yeah, I want to see if the tests pass on jenkins

masterleinad:
With the relaxed tolerances, the tests pass at least locally both with `master` and `fastap_mpi`.

tests/test_hierarchy_device.cu
```diff
@@ -426,7 +426,7 @@ BOOST_AUTO_TEST_CASE(amgx)
     // We only supports Jacobi smoother on the device
     params->put("smoother.type", "Jacobi");
 
-    double const tolerance = 1.;
+    double const tolerance = 5.;
```
dalg24: Please call it tolerance_percent.  I jumped when I saw it initially when I thought it might be relative tolerance

dalg24: Sure

PR #137, opened 2019-03-11
Down with matrices!

dalg24:
Summary of changes: less matrices/more marix-free

`DealIIMatrixFreeMeshEvaluator` still derives from `DealIIMeshEvaluator` so that we can defer to the existing user customization point `DealIIMeshEvaluator::{evaluate_global(dof_handler, constraints, matrix), evaluate_agglomerate(dof_handler, constraints, sparsity_pattern, matrix)}` in `DealIIMatrixFreeMeshEvaluator::{matrix_free_evaluate_global(src, dst), matrix_free_evaluate_agglomerate(dof_handler, src, dst)}` but note that the signatures of the member functions are different.

Also added `DealIIMatrixFreeMeshEvaluator::{matrix_free_get_agglomerate_diagonal(dof_handler), matrix_free_get_diagonal_inverse()}` and actually removed the matrices in the class everywhere except the body of the future new customization point that are calling the old ones for now.

PR #139, opened 2019-03-12
Revert changes that disable all tests but the hierarchy ones

dalg24:
I committed these changes by inadvertence in #137 :(


dalg24:
For reference `git commit -a` accident was d6f293c

PR #144, opened 2019-03-21
Allow changing FE degree from input file

### LINKED ISSUE #142: Ease changing FE degree in tests
Right now, setting FE degree has a drawback: one has to remember to change it in two places, `Laplace` constructor and `MeshEvaluator`. Instead, it should be done through a parameter list that is passed to both.

aprokop:
Fix #142.

### DIFF
--- a/tests/hierarchy_driver.cc
+++ b/tests/hierarchy_driver.cc
@@ -21,7 +21,8 @@ void main_(std::shared_ptr<boost::property_tree::ptree> params)
   Source<dim> source;
 
   auto laplace_ptree = params->get_child("laplace");
-  Laplace<dim, DVector> laplace(comm, 1);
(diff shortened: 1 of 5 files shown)


Rombur:
`/var/jenkins/workspace/mfmg-experimental/tests/test_hierarchy_device.cu(291): error: no instance of constructor "TestMeshEvaluator<dim>::TestMeshEvaluator [with dim=2]" matches the argument list
            argument types are: (MPI_Comm, dealii::DoFHandler<2, 2>, dealii::AffineConstraints<double>, unsigned int, const dealii::TrilinosWrappers::SparseMatrix, std::shared_ptr<dealii::Function<2, double>>, mfmg::CudaHandle)
          detected during instantiation of "double test<dim>(std::shared_ptr<boost::property_tree::ptree>) [with dim=2]" `

aprokop:
Fixed.

PR #145, opened 2019-03-25
Fix a bug where Lanczos was not using the user's initial guess

Rombur:
Not ready for review. I have updated my version of deal.II and it's not compatible with mfmg. So I don't know if the tests pass.

### DIFF
--- a/include/mfmg/dealii/amge_host.templates.hpp
+++ b/include/mfmg/dealii/amge_host.templates.hpp
@@ -45,8 +45,10 @@ struct MatrixFreeAgglomerateOperator
 
   template <typename DoFHandler>
   MatrixFreeAgglomerateOperator(MeshEvaluator const &mesh_evaluator,
-                                DoFHandler &dof_handler)
-      : _mesh_evaluator{mesh_evaluator}, _dof_handler{dof_handler}
+                                DoFHandler &dof_handler,
+                                dealii::AffineConstraints<double> &constraints)
+      : _mesh_evaluator(mesh_evaluator), _dof_handler(dof_handler),
+        _constraints(constraints)
   {
   }
 
@@ -58,7 +60,8 @@ struct MatrixFreeAgglomerateOperator
 
   std::vector<double> get_diag_elements() const
   {
-    return _mesh_evaluator.matrix_free_get_agglomerate_diagonal(_dof_handler);
+    return _mesh_evaluator.matrix_free_get_agglomerate_diagonal(_dof_handler,
+                                                                _constraints);
   }
 
   size_type m() const { return _dof_handler.n_dofs(); }
@@ -69,6 +72,7 @@ private:
   MeshEvaluator const &_mesh_evaluator;
   static int constexpr dim = MeshEvaluator::_dim;
   dealii::DoFHandler<dim> &_dof_handler;
+  dealii::AffineConstraints<double> &_constraints;
 };
 
 template <int dim, typename MeshEvaluator, typename VectorType>
@@ -87,6 +91,7 @@ void lanczos_compute_eigenvalues_and_eigenvectors(
     unsigned int n_eigenvectors, double tolerance,
     boost::property_tree::ptree const &eigensolver_params,
     AgglomerateOperator const &agglomerate_operator,
+    dealii::Vector<double> const &initial_guess,
     std::vector<std::complex<double>> &eigenvalues,
     std::vector<dealii::Vector<double>> &eigenvectors)
 {
@@ -116,7 +121,8 @@ void lanczos_compute_eigenvalues_and_eigenvectors(
       agglomerate_operator);
 
   std::vector<double> real_eigenvalues;
-  std::tie(real_eigenvalues, eigenvectors) = solver.solve(lanczos_params);
+  std::tie(real_eigenvalues, eigenvectors) =
+      solver.solve(lanczos_params, initial_guess);
   ASSERT(n_eigenvectors == eigenvectors.size(),
          "Wrong number of computed eigenpairs");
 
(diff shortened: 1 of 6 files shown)


Rombur:
Looking back we were happy that `lapack` and `Lanczos` gave the same results but we know that `lapack` contains the spurious eigenvalues since there is no way to get ride of them. This should have been a red flag but we missed it :(

Rombur:
OK for review. The interface is not that great right now but I will change it in an upcoming PR.

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -160,9 +168,11 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
       _eigensolver_params.get<std::string>("type", "lanczos");
   if (eigensolver_type == "lanczos")
   {
+    dealii::Vector<double> initial_vector(n_dofs_agglomerate);
```
aprokop: Move these lines out of if-else branches, all eigensolvers need them.

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -260,9 +270,11 @@ AMGe_host<dim, MeshEvaluator, VectorType>::compute_local_eigenvectors(
   }
   else if (eigensolver_type == "lanczos")
   {
+    dealii::Vector<double> initial_vector(n_dofs_agglomerate);
+    evaluator.set_initial_guess(agglomerate_constraints, initial_vector);
```
aprokop: Remove these, see above.

include/mfmg/dealii/lanczos.hpp
```diff
@@ -35,7 +35,8 @@ class Lanczos
 
   // Operations
   std::tuple<std::vector<double>, std::vector<VectorType>>
-  solve(boost::property_tree::ptree const &params) const;
+  solve(boost::property_tree::ptree const &params,
+        VectorType initial_guess) const;
```
aprokop: I think initial guess should be optional.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -351,15 +351,12 @@ template <typename OperatorType, typename VectorType>
...
-  // of PDE problems
-  initial_guess = 1.;
-
   // Add random noise to the guess
   std::mt19937 gen(seed);
   std::uniform_real_distribution<double> dist(0, 1);
   std::transform(initial_guess.begin(), initial_guess.end(),
-                 initial_guess.begin(), [&](auto &v) { return v + dist(gen); });
+                 initial_guess.begin(),
+                 [&](auto &v) { return (1. + dist(gen)) * v; });
```
aprokop: Need an extensive comment here about the care that went into this very fragile line. Specifically, it hides the idea that it preserves the zeros in the initial guess that are already there, which correspond to constraints.

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -66,10 +66,10 @@ Lanczos<OperatorType, VectorType>::solve(
   DeflatedOperator<OperatorType, VectorType> deflated_op(_op);
 
   // Loop over Lanczos solves
-  VectorType initial_guess(_op.n());
   for (int cycle = 0; cycle < num_cycles; ++cycle)
   {
-    details_set_initial_guess(initial_guess, cycle);
+    if (cycle > 0)
```
aprokop: I disagree with this. It would make sense not to rely on a caller to provide a randomly perturbed initial guess. I think `details_set_initial_guess` should still be called on the user provided guess.

include/mfmg/dealii/amge_host.templates.hpp
Rombur: Well we support only one eigensolver...

include/mfmg/dealii/lanczos.hpp
Rombur: I thought about it but I don't see when we would want to use it except in the tests.

include/mfmg/dealii/lanczos.templates.hpp
Rombur: That's what I did at first but I fear than in three months we will want to use a particular initial guess and we will have forgotten that we don't use the user-provided initial guess. This will be a pain to debug. This is also consistent with what we do with arpack.

include/mfmg/dealii/amge_host.templates.hpp
aprokop: Fair point for matrix-free, but with matrices we still do the same thing. I would just always have initial guess out of if-else, and this way both versions would look the same.

include/mfmg/dealii/lanczos.templates.hpp
aprokop: Fair point, I have no counter argument.

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -69,6 +72,7 @@ struct MatrixFreeAgglomerateOperator
   MeshEvaluator const &_mesh_evaluator;
   static int constexpr dim = MeshEvaluator::_dim;
   dealii::DoFHandler<dim> &_dof_handler;
+  dealii::AffineConstraints<double> &_agglomerate_constraints;
```
dalg24: s/_agglomerate_constraints/_constraints for consistency

include/mfmg/dealii/lanczos.templates.hpp
dalg24: Add a comment to capture this discussion.

Rombur:
Done. Sorry I squashed and forced-push right away...

include/mfmg/dealii/lanczos.templates.hpp
```diff
@@ -351,15 +351,14 @@ template <typename OperatorType, typename VectorType>
...
     VectorType &initial_guess, int seed)
 {
-  // Set initial guess to constant vector to try to capture "smooth" eigenmodes
-  // of PDE problems
-  initial_guess = 1.;
-
-  // Add random noise to the guess
+  // Add random noise to the guess. Each element of the initial guess is scale
+  // by a random non-zero factor. This ensures that entries that have been set
+  // to zero because they are associated with constrained dofs stay nul.
```
dalg24: s/nul/null/

aprokop:
```
Modify initial guess with a random noise by multiplying each entry of the vector with a random value from a uniform distribution. This specific procedure guarantees that zero entries of the vector stay zero, which is important for situations where they are associated with constrained dofs in DealII.
```

PR #146, opened 2019-04-02
Use matrix-free mesh evaluator

Rombur:
This is not ready to be reviewed it is just to share with @aprokop. @aprokop you should care about the new test `benchmark_mf` It's the same as the matrix-based benchmark

### DIFF
--- a/include/mfmg/common/amge.templates.hpp
+++ b/include/mfmg/common/amge.templates.hpp
@@ -496,7 +496,7 @@ template <int dim, typename VectorType>
 unsigned int AMGe<dim, VectorType>::build_agglomerates_block(
     std::array<unsigned int, dim> const &agglomerate_dim) const
 {
-  // Faces in deal.II are orderd as follows: left (x_m) = 0, right (x_p) = 1,
+  // Faces in deal.II are ordered as follows: left (x_m) = 0, right (x_p) = 1,
   // front (y_m) = 2, back (y_p) = 3, bottom (z_m) = 4, top (z_p) = 5
   unsigned int constexpr x_p = 1;
   unsigned int constexpr y_p = 3;
--- a/include/mfmg/common/amge.templates.hpp
+++ b/include/mfmg/common/amge.templates.hpp
@@ -14,6 +14,7 @@
 
 #include <mfmg/common/amge.hpp>
 #include <mfmg/common/exceptions.hpp>
+#include <mfmg/common/utils.hpp>
 
 #include <deal.II/distributed/tria.h>
 #include <deal.II/dofs/dof_accessor.h>
@@ -314,63 +315,6 @@ void AMGe<dim, VectorType>::compute_restriction_sparse_matrix(
 
   // Compress the matrix
   restriction_sparse_matrix.compress(dealii::VectorOperation::add);
-
-#if MFMG_DEBUG
-  // Check that the locally_relevant_global_diag is the sum of the agglomerates
-  // diagonal
-  // TODO do not ask user for the locally_relevant_global_diag
-  dealii::LinearAlgebra::distributed::Vector<typename VectorType::value_type>
-      new_global_diag(locally_relevant_global_diag.get_partitioner());
-  for (unsigned int i = 0; i < n_agglomerates; ++i)
-  {
-    // Get the size of the eigenvectors in agglomerate i
-    unsigned int offset = std::accumulate(n_local_eigenvectors.begin(),
-                                          n_local_eigenvectors.begin() + i, 0);
-    unsigned int const n_elem = eigenvectors[offset].size();
-
-    for (unsigned int j = 0; j < n_elem; ++j)
-    {
-      dealii::types::global_dof_index const global_pos = dof_indices_maps[i][j];
-      new_global_diag[global_pos] += diag_elements[i][j];
-    }
-  }
-  new_global_diag.compress(dealii::VectorOperation::add);
-  new_global_diag -= locally_relevant_global_diag;
-  ASSERT((new_global_diag.linfty_norm() /
-          locally_relevant_global_diag.linfty_norm()) < 1e-14,
-         "Sum of agglomerate diagonals is not equal to the global diagonal");
-
-  // Check that the sum of the weight matrices is the identity
-  auto locally_owned_dofs =
-      locally_relevant_global_diag.locally_owned_elements();
-  dealii::TrilinosWrappers::SparsityPattern sp(locally_owned_dofs,
-                                               locally_owned_dofs, this->_comm);
-  for (auto local_index : locally_owned_dofs)
-    sp.add(local_index, local_index);
-  sp.compress();
-
-  dealii::TrilinosWrappers::SparseMatrix weight_matrix(sp);
-  pos = 0;
-  for (unsigned int i = 0; i < n_agglomerates; ++i)
-  {
-    unsigned int const n_elem = eigenvectors[pos].size();
-    for (unsigned int j = 0; j < n_elem; ++j)
-    {
-      dealii::types::global_dof_index const global_pos = dof_indices_maps[i][j];
-      double const value =
-          diag_elements[i][j] / locally_relevant_global_diag[global_pos];
-      weight_matrix.add(global_pos, global_pos, value);
-    }
-    pos += n_local_eigenvectors[i];
-  }
-
-  // Compress the matrix
-  weight_matrix.compress(dealii::VectorOperation::add);
-
-  for (auto index : locally_owned_dofs)
-    ASSERT(std::abs(weight_matrix.diag_element(index) - 1.0) < 1e-14,
-           "Sum of local weight matrices is not the identity");
-#endif
 }
 
 template <int dim, typename VectorType>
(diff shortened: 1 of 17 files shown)


Rombur:
Ready for review

Rombur:
_ping_

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -35,25 +35,22 @@ std::shared_ptr<Operator<VectorType>>
 DealIIHierarchyHelpers<dim, VectorType>::get_global_operator(
     std::shared_ptr<MeshEvaluator> mesh_evaluator)
 {
-  if (_global_operator == nullptr)
```
masterleinad: How did this code create a memory leak?

For the new implementation `get_global_operator` sounds a little like a misnomer. I would have expected the former behavior for something called `get` (and probably also `const`). The new one rather creates the operator, doesn't it?

masterleinad: So something like https://godbolt.org/z/Vhnizg.

Rombur: yeah something like that

include/mfmg/common/amge.templates.hpp
```diff
@@ -496,7 +496,7 @@ template <int dim, typename VectorType>
 unsigned int AMGe<dim, VectorType>::build_agglomerates_block(
     std::array<unsigned int, dim> const &agglomerate_dim) const
 {
-  // Faces in deal.II are orderd as follows: left (x_m) = 0, right (x_p) = 1,
+  // Faces in deal.II are ordered as follows: left (x_m) = 0, right (x_p) = 1,
```
masterleinad: Looks good!

source/dealii/dealii_hierarchy_helpers.cc
Rombur: I don't have a strong opinion on the name but technically we are not building the global operator because it should already exists in the user's code (this is the system the user wants to solver). So we are getting it from the user but I am open to change the name.

include/mfmg/cuda/amge_device.templates.cuh
```diff
@@ -339,6 +340,9 @@ AMGe_device<dim, MeshEvaluator, VectorType>::compute_restriction_sparse_matrix(
   AMGe<dim, VectorType>::compute_restriction_sparse_matrix(
       eigenvectors, diag_elements, dof_indices_maps, n_local_eigenvectors,
       locally_relevant_global_diag, restriction_sparse_matrix);
+  check_restriction_matrix(this->_comm, eigenvectors, dof_indices_maps,
```
masterleinad: This commit looks good as well. You are just moving stuff.

source/dealii/dealii_matrix_free_operator.cc
```diff
@@ -105,11 +105,9 @@ DealIIMatrixFreeOperator<dim, VectorType>::multiply_transpose(
   auto tmp = this->build_range_vector();
   auto b_mat = downcast_b->get_matrix();
 
-  auto c_mat = std::make_shared<dealii::TrilinosWrappers::SparseMatrix>(
+  auto c_mat = matrix_transpose_matrix_multiply(
```
masterleinad: This commit looks reasonable as well but I didn't look at the commented part.

source/dealii/dealii_matrix_free_smoother.cc
```diff
@@ -50,9 +50,7 @@ DealIIMatrixFreeSmoother<dim, VectorType>::DealIIMatrixFreeSmoother(
       data.max_eigenvalue = *max_eigenvalue;
     }
 
-    auto diagonal_inverse = std::make_shared<preconditioner_type>();
-    diagonal_inverse->reinit(matrix_free_operator->get_diagonal_inverse());
-    data.preconditioner = diagonal_inverse;
+    data.preconditioner = matrix_free_operator->get_diagonal_inverse();
```
masterleinad: Looks reasonable as well.

tests/laplace_matrix_free.hpp
```diff
@@ -257,7 +263,7 @@ void LaplaceMatrixFree<dim, fe_degree, ScalarType>::setup_system(
   }
 
   if (ptree.get("distort_random", false))
-    dealii::GridTools::distort_random(0.2, _triangulation);
+    dealii::GridTools::distort_random(0.1, _triangulation);
```
masterleinad: Any reason for changing this?

Rombur: Oh yeah this was a bug. We were using `0.1` for the matrix-based so I just changed to the same value for the matrix-free.

tests/test_hierarchy.cc
```diff
@@ -32,12 +32,88 @@
...
+  dealii::ConditionalOStream pcout(
+      std::cout, dealii::Utilities::MPI::this_mpi_process(comm) == 0);
+
+  auto material_property =
+      MaterialPropertyFactory<dim>::create_material_property(
+          params->get<std::string>("material_property.type"));
+  Source<dim> source;
+
+  auto laplace_ptree = params->get_child("laplace");
+  int constexpr fe_degree = 1;
```
masterleinad: All power to `east const`! :wink:

tests/test_hierarchy.cc
```diff
@@ -32,12 +32,88 @@
...
+  LaplaceMatrixFree<dim, fe_degree, double> mf_laplace(comm);
+  mf_laplace.setup_system(laplace_ptree, *material_property);
+
+  auto const locally_owned_dofs = mf_laplace._locally_owned_dofs;
+  DVector solution(locally_owned_dofs, comm);
+  DVector rhs(mf_laplace._system_rhs);
```
masterleinad: I don't really like this style, but we use it everywhere already.

tests/test_hierarchy_helpers.hpp
```diff
@@ -50,11 +51,30 @@ class Source final : public dealii::Function<dim>
...
+  virtual dealii::VectorizedArray<double>
+  value(dealii::Point<dim, dealii::VectorizedArray<double>> const &p,
+        unsigned int const = 0) const = 0;
+
+  virtual double value(dealii::Point<dim> const &p,
+                       unsigned int const component = 0) const override = 0;
+};
+
+template <int dim>
+class ConstantMaterialProperty : public Coefficient<dim>
```
masterleinad: Can't this class be `final` again?

tests/test_hierarchy_helpers.hpp
```diff
@@ -63,11 +83,19 @@ class ConstantMaterialProperty final : public dealii::Function<dim>
 };
 
 template <int dim>
-class LinearXMaterialProperty final : public dealii::Function<dim>
+class LinearXMaterialProperty : public Coefficient<dim>
```
masterleinad: Can't this class be `final` again?

tests/test_hierarchy_helpers.hpp
```diff
@@ -76,28 +104,52 @@ class LinearXMaterialProperty final : public dealii::Function<dim>
 };
 
 template <int dim>
-class LinearMaterialProperty : public dealii::Function<dim>
+class LinearMaterialProperty : public Coefficient<dim>
```
masterleinad: Can't this class be `final` again?

tests/test_hierarchy_helpers.hpp
```diff
@@ -50,11 +51,30 @@ class Source final : public dealii::Function<dim>
...
-class ConstantMaterialProperty final : public dealii::Function<dim>
+class Coefficient : public dealii::Function<dim>
+{
+public:
+  virtual dealii::VectorizedArray<double>
+  value(dealii::Point<dim, dealii::VectorizedArray<double>> const &p,
+        unsigned int const = 0) const = 0;
+
+  virtual double value(dealii::Point<dim> const &p,
+                       unsigned int const component = 0) const override = 0;
```
masterleinad: You don't really need this member function here, do you?

tests/test_hierarchy_helpers.hpp
```diff
@@ -50,11 +51,30 @@ class Source final : public dealii::Function<dim>
 };
 
 template <int dim>
-class ConstantMaterialProperty final : public dealii::Function<dim>
+class Coefficient : public dealii::Function<dim>
```
masterleinad: Isn't this more a `VectorizedFuntion` class or similar?

tests/test_hierarchy_helpers.hpp
```diff
@@ -50,11 +51,30 @@ class Source final : public dealii::Function<dim>
...
+{
+public:
+  virtual dealii::VectorizedArray<double>
+  value(dealii::Point<dim, dealii::VectorizedArray<double>> const &p,
+        unsigned int const = 0) const = 0;
```
masterleinad: Where is this function actually used?

tests/test_hierarchy_helpers.hpp
Rombur: Yeah I work on too many branches at the same time :/

tests/test_hierarchy_helpers.hpp
Rombur: Not really. The class is used for matrix-free where it is vectorized and the matrix-based which is not vectorized.

tests/test_hierarchy_helpers.hpp
Rombur: https://github.com/ORNL-CEES/mfmg/blob/master/tests/laplace_matrix_free.hpp#L109

tests/test_hierarchy_helpers.hpp
Rombur: I do for the matrix-based

tests/test_hierarchy_helpers.hpp
masterleinad: Thanks!

tests/test_hierarchy_helpers.hpp
masterleinad: What I was trying to say is that the base class already defines this function as pure virtual and we could omit the definition here.
If you want to keep it for documentation purposes/helping to understand the code, it's fine with me to keep it, though.

Rombur: Oh I see. Yes, I think it's easier to see what is necessary to implement if everything is here. Otherwise, people who don't know `dealii::Function` won't know that they have to implement it.

PR #147, opened 2019-04-03
Fix compiling with CUDA_MPI=ON

masterleinad:
The signature for `MPI_Comm_size` is
```
int MPI_Comm_size(MPI_Comm comm, int *size)
```


Rombur:
OK to test

dalg24:
OK to test

Rombur:
> How the heck was it never caught?

Because we don't have cuda mpi on the tester so the code is not tested

PR #149, opened 2019-04-03
Allow finding TPLs by environment variables

masterleinad:
Also provide an error message in case `AmgX` was not found.


cmake/SetupTPLs.cmake
```diff
@@ -36,12 +36,23 @@ ENDIF()
...
+      HINTS ${AMGX_DIR} $ENV{AMGX_DIR}
+      PATH_SUFFIXES lib)
+    FIND_PATH(AMGX_INCLUDE_DIR amgx_c.h
+      HINTS ${AMGX_DIR} $ENV{AMGX_DIR}
+      PATH_SUFFIXES include)
+
+    IF(${AMGX_INCLUDE_DIR} MATCHES "-NOTFOUND" OR ${AMGX_LIBRARY} MATCHES "-NOTFOUND")
+      MESSAGE(FATAL_ERROR 
+              "Error! Could not locate AmgX.")
+    ENDIF()
```
dalg24: Do this instead
```diff
-  IF(${AMGX_INCLUDE_DIR} MATCHES "-NOTFOUND" OR ${AMGX_LIBRARY} MATCHES "-NOTFOUND")
-    MESSAGE(FATAL_ERROR "Error! Could not locate AmgX.")
-  ENDIF()
+  find_package_handle_standard_args(AmgX DEFAULT_MSG AMGX_INCLUDE_DIR AMGX_LIBRARY)
```

dalg24: We found that the macro `find_package_handle_standard_args` would only make sense if we wrote a module to find AmgX and decided to change to `IF(NOT AMGX_INCLUDE_DIR OR NOT AMGX_LIBRARY)`

dalg24:
> Let's not merge this. The build system and this part in particular is very brittle and it is too important to start making change to it at this time in the project. Passing the jenkins is not enough, we need to try on different machines.

This patch is armless and gives cleaner code in CMake.

dalg24:
We can strip the part with the env variables if you don't like but this is the correct way to detect the AmgX lib and link against it.

Rombur:
We shouldn't merge anything in the build system until we know it works on Docker, my machine, and Andrey's.

Issue #150, opened 2019-04-05
LAPACK tests not working


masterleinad:
After updating `deal.II`, i.e. after https://github.com/dealii/dealii/pull/7687, the `LAPACK` tests in `test_hierarchy` were not working (#138). One of the reasons is that the agglomerates might not have boundary conditions and hence the corresponding matrix has a non-trivial kernel shifting the diagonal entries slightly at least allows inverting the matrix (using `UMFPACK`). I observed this even before updating, i.e. with with the version in the image. After that the `LAPACK` tests are still failing due to the restriction matrix not being full-rank (possibly due to the spurious eigenvalues).
Just switching from `LAPACK` to `ARPACK` gives similar results as with `Lanczos`. Copying the reference values from the `Lanczos` tests to the `ARPACK` tests gives:

```
1: /tmp/mfmg/tests/test_hierarchy.cc(241): error: in "hierarchy_3d/_16": check conv_rate == ref_solution[std::make_tuple(mesh, distort_random_str, reordering, eigensolver, matrix_free_str)] has failed [0.11466298397904537 != 0.1148067738]. Relative difference exceeds tolerance [0.00125402 > 1e-06]
1: Failure occurred in a following context:
1:     mesh = hyper_ball; distort_random = false; reordering = None; mesh_evaluator_type = DealIIMeshEvaluator; eigensolver = arpack; 
1: FE degree: 1
1: Convergence rate: 0.11
1: /tmp/mfmg/tests/test_hierarchy.cc(241): error: in "hierarchy_3d/_17": check conv_rate == ref_solution[std::make_tuple(mesh, distort_random_str, reordering, eigensolver, matrix_free_str)] has failed [0.11469510272606935 != 0.1148067738]. Relative difference exceeds tolerance [0.000973634 > 1e-06]
1: Failure occurred in a following context:
1:     mesh = hyper_ball; distort_random = false; reordering = None; mesh_evaluator_type = DealIIMeshEvaluator; eigensolver = lanczos; 
1: skip
1: FE degree: 1
1: Convergence rate: 0.30
1: /tmp/mfmg/tests/test_hierarchy.cc(241): error: in "hierarchy_3d/_19": check conv_rate == ref_solution[std::make_tuple(mesh, distort_random_str, reordering, eigensolver, matrix_free_str)] has failed [0.2995474407510747 != 0.3012330587]. Relative difference exceeds tolerance [0.00562722 > 1e-06]
1: Failure occurred in a following context:
[...]

PR #152, opened 2019-04-08
Shift eigenvalues for the agglomerate matrices to make them invertible

masterleinad:
Partially addresses #150 and #138. We know that the agglomerate matrices might not be invertible due to missing boundary conditions. Since we expect the underlying operator to be positive definite shifting the eigenvalues by one fixes this issue. In the end, we don't care for the eigenvalues but only for the eigenvectors for building the restriction matrix.


aprokop:
I think I'm missing something here. Shifting diagonal is not the same as shifting eigenvalues.

masterleinad:
Locally this passes for me.

aprokop:
@masterleinad Please don't force-push opened PR, just commit on top of it.

Rombur:
Why did you open two PR? I don't see the difference with #153

PR #153, opened 2019-04-08
Test replace arpack lapack

masterleinad:
As we have seen in #150 and #138, `LAPACK` is not reliable when building the restriction matrix.
Hence, test `ARPACK` instead.


masterleinad:
Based on #152.

Rombur:
can you rebase this on master

masterleinad:
Sure, done.

PR #158, opened 2019-04-10
Revert "Find LAPACKE and link to it"

Rombur:
This commit silently breaks the CI. Reverting it since we can't run/test the code with it.


masterleinad:
Let me try another solution first.

PR #159, opened 2019-04-10
Fix CI

masterleinad:
This PR makes sure that we fail if the `cmake` call isn't successfull. Furthermore, we first try if the `LAPACK` interface provided by `deal.II` is sufficient for `LAPACKE` and only search for an additional library if that fails.

Alternative to #158.

### DIFF
--- a/ci/compile_and_run.sh
+++ b/ci/compile_and_run.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
-cd $1
+cd $1 || exit 1
 rm -rf build
-mkdir build && cd build
+mkdir build && cd build || exit 1
 ARGS=(
   -D CMAKE_BUILD_TYPE=Debug
   -D MFMG_ENABLE_TESTS=ON
@@ -15,7 +15,7 @@ ARGS=(
   -D AMGX_DIR=${AMGX_DIR}
   -D CMAKE_CXX_FLAGS="-Wall -Wpedantic -Wextra -Wshadow"
   )
-cmake "${ARGS[@]}" ../
+cmake "${ARGS[@]}" ../ || exit 1
 make -j12
 # Because Arpack is not thread-safe we cannot use multithreading
 export DEAL_II_NUM_THREADS=1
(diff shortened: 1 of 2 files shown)


ci/compile_and_run.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
-cd $1
+cd $1 || exit 1
```
dalg24: Why not just `set -e`?

cmake/SetupTPLs.cmake
```diff
@@ -39,18 +39,35 @@ ELSE()
...
-FIND_LIBRARY(LAPACKE_LIBRARY NAMES lapacke
-             HINTS ${LAPACK_DIR} $ENV{LAPACK_DIR}
-             PATH_SUFFIXES lib)
 FIND_PATH(LAPACKE_INCLUDE_DIR lapacke.h
           HINTS ${LAPACK_DIR} $ENV{LAPACK_DIR}
           PATH_SUFFIXES include)
 
-IF(NOT LAPACKE_INCLUDE_DIR OR NOT LAPACKE_LIBRARY)
+IF(NOT LAPACKE_INCLUDE_DIR)
+  MESSAGE(STATUS "${LAPACKE_INCLUDE_DIR}")
```
dalg24: What is the point of printing it if it was not found?

cmake/SetupTPLs.cmake
```diff
@@ -39,18 +39,35 @@ ELSE()
...
 ENDIF()
 
+# First test if the libraries deal.II links to already include LAPACKE
+INCLUDE(CheckFunctionExists)
+SET(CMAKE_REQUIRED_LIBRARIES "${DEAL_II_LIBRARIES}")
+CHECK_FUNCTION_EXISTS(LAPACKE_dgeqrf LAPACKE_WORKS)
+
+# If not try to find a separate library
+IF (LAPACKE_WORKS)
+  SET(LAPACKE_LIBRARY "")
```
dalg24: Do we need to define the variable at all?

PR #164, opened 2019-04-17
Improve CMake II

masterleinad:
This is a rebased version of #82. I'd just like to see if there are still problems.
Locally, everything looks fine.


cmake/SetupCUDA.cmake
```diff
@@ -2,23 +2,39 @@
 # the list of CUDA libraries that we are using
 FUNCTION(SET_CUDA_LIBRARIES)
   ADD_DEFINITIONS(-DMFMG_WITH_CUDA)
+  FIND_LIBRARY(CUSPARSE cusparse PATH ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+  FIND_LIBRARY(CUSOLVER cusolver PATH ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+  MESSAGE("implicit ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}")
```
dalg24: Did you mean to leave that print statement?

cmake/SetupCUDA.cmake
```diff
@@ -2,23 +2,39 @@
...
+    IF (${MFMG_ENABLE_AMGX})
+      FIND_LIBRARY(AMGX_LIBRARY amgx amgxsh PATH "${AMGX_DIR}/lib")
+      FIND_LIBRARY(CUBLAS cublas HINT ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+      SET(AMGX_LIBRARIES
+        ${AMGX_LIBRARY}
+        ${CUBLAS}
+        PARENT_SCOPE
+        )
+      SET(AMGX_INCLUDE_DIR "${AMGX_DIR}/include" PARENT_SCOPE)
+      ADD_DEFINITIONS(-DMFMG_WITH_AMGX)
```
dalg24: NOOO.  Add it to the target.

ci/compile_and_run.sh
```diff
@@ -1,8 +1,20 @@
...
+# toolkit is installed. The stub libraries allow you to compile the code but not
+# to run it because the symbol are missing. The idea is that at run time, we will
+# pick the correct libraries instead of the stub ones. This doesn't work when
+# using an `rpath` (see https://github.com/NVIDIA/nvidia-docker/issues/775). For
+# some reason the problem only appears with CUSolver having missing symbol to
+# OpenMP. Removing the stubs fixes the problem.
+rm -r  /usr/local/cuda/lib64/stubs
 cd $1 || exit 1
 rm -rf build
+export LD_LIBRARY_PATH=/usr/lib/gcc/x86_64-linux-gnu/5.4.0:${LD_LIBRARY_PATH}
```
dalg24: Why are you setting this environment variable?

cmake/SetupCUDA.cmake
```diff
@@ -2,23 +2,39 @@
...
+  FIND_LIBRARY(CUSPARSE cusparse PATH ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+  FIND_LIBRARY(CUSOLVER cusolver PATH ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+  MESSAGE("implicit ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}")
   SET(MFMG_CUDA_LIBRARIES
-    "cusparse"
-    "cusolver"
+    ${CUSPARSE}
+    ${CUSOLVER}
     PARENT_SCOPE
     )
```
dalg24: @Rombur It feels really wrong to have to set in the parent scope without even checking that we actually found the libraries or/and the include directories.  What was the rational to not define modules?  If we were to do so, we would be able to call `find_package_handle_standard_args()` and crash at configure time if `REQUIRED` was specified.

Rombur:
This is a rebase of WIP PR. I made it work on my machine and I've hacked docker so that it will work but that's not the real problem.

cmake/SetupCUDA.cmake
masterleinad: I just rebased and didn't touch anything yet.

PR #166, opened 2019-04-24
Remove unused variable

Rombur:
Silence warnings


masterleinad:
We might just run CI with `-Werror`.

PR #167, opened 2019-04-24
Improve CI script

masterleinad:
Apply a suggestion in  https://github.com/ORNL-CEES/mfmg/pull/159#discussion_r275197426 and run ci with `-Werror` so we don't forget to address warnings in pull requests.

### DIFF
--- a/ci/compile_and_run.sh
+++ b/ci/compile_and_run.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
-cd $1 || exit 1
+set -e
+cd $1
 rm -rf build
-mkdir build && cd build || exit 1
+mkdir build && cd build
 ARGS=(
   -D CMAKE_BUILD_TYPE=Debug
   -D MFMG_ENABLE_TESTS=ON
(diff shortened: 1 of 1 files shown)


ci/compile_and_run.sh
```diff
@@ -13,10 +14,10 @@ ARGS=(
   -D DEAL_II_DIR=${DEAL_II_DIR}
   -D MFMG_ENABLE_AMGX=ON
   -D AMGX_DIR=${AMGX_DIR}
-  -D CMAKE_CXX_FLAGS="-Wall -Wpedantic -Wextra -Wshadow"
+  -D CMAKE_CXX_FLAGS="-Wall -Wpedantic -Wextra -Wshadow -Werror"
```
dalg24: Is that really a good idea?

masterleinad: What are your objections?

dalg24: Won't even run the tests if you have a warning that you might have missed cause you use a different compiler

masterleinad: For me, `CI` should check for all the requirements we want a pull request/the code to fulfill in the end and not so much as a development tool when creating a pull request.

In `deal.II`, we just have multiple checks were one would one would not have `-Werror` and run tests while another one would check, e.g. for indentation (not as a test) and for compiler warnings.

I would also be in favor to run `clang-tidy` if we want to enforce its checks in the end.

ci/compile_and_run.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
-cd $1 || exit 1
+set -e
```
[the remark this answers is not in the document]
Rombur: done

ci/compile_and_run.sh
Rombur: What do you guys want to do here? I don't have a strong opinion either way

ci/compile_and_run.sh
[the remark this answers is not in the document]
masterleinad: I am using this script for testing pull requests locally and would like to keep this option (or go with the previous version).

PR #168, opened 2019-04-24
Address empty patches

masterleinad:
We need to make sure that we can deal with the case that MPI processes don't own any patch.

### DIFF
--- a/source/dealii/dealii_hierarchy_helpers.cc
+++ b/source/dealii/dealii_hierarchy_helpers.cc
@@ -111,10 +111,15 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
                        boost::hash<std::pair<unsigned int, unsigned int>>>
         delta_correction_acc;
     bool is_halo_agglomerate = false;
+
+    // In case there are no patches we own, we still need to construct the
+    // restriction operators. We set the number of eigenvalues to zero instead
+    // of returning early and duplicating the setup.
     unsigned int const n_local_eigenvectors =
         interior_agglomerates.empty()
             ? 0
             : delta_correction_matrix.m() / interior_agglomerates.size();
+
     for (auto const &agglomerates_vector :
          {interior_agglomerates, halo_agglomerates})
     {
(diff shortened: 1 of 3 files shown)


Rombur:
That just looks wrong to me. How is it possible that a process doesn't have any agglomerate? The only case where it is not a bug it's if the process doesn't own any cells.

masterleinad:
> The only case where it is not a bug it's if the process doesn't own any cells.

Yes, exactly. For me that happens with one global refinement and two MPI processes.

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -112,7 +112,9 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
         delta_correction_acc;
     bool is_halo_agglomerate = false;
     unsigned int const n_local_eigenvectors =
-        delta_correction_matrix.m() / interior_agglomerates.size();
+        interior_agglomerates.empty()
+            ? 0
+            : delta_correction_matrix.m() / interior_agglomerates.size();
```
Rombur: Shouldn't we just return at this point

source/dealii/dealii_matrix_free_hierarchy_helpers.cc
```diff
@@ -109,7 +109,9 @@ DealIIMatrixFreeHierarchyHelpers<dim, VectorType>::build_restrictor(
         delta_correction_acc;
     bool is_halo_agglomerate = false;
     unsigned int const n_local_eigenvectors =
-        delta_correction_matrix.m() / interior_agglomerates.size();
+        interior_agglomerates.empty()
+            ? 0
+            : delta_correction_matrix.m() / interior_agglomerates.size();
```
Rombur: Same here

aprokop:
Could a better approach be to do a comm split on the finest level, and just work with MPI processes that have at least one cell?

masterleinad:
I am not sure how much work that is.

Rombur:
Unless you really want to change the communicator @aprokop, I think we can merge the current PR

PR #169, opened 2019-04-29
Fast AP:  Only access DoFs contained in the agglomerates 

masterleinad:
Without this pull request we try to access entries in `delta_eigenvector_matrix` and `eigenvector_matrix`
that are not corresponding to the current agglomerate (and consequently don't exist). Thus, using `operator()` instead of `el()` gives errors. `el()` would just ignore any entries that don't exist and hence worked fine. However, it is hard to debug the code, e.g. for `MPI`, if `operator()` can not be used.

While the changes look large, we really only address an interior agglomerate and its halo agglomerate at the same time. Hence most code is just duplicated for the two agglomerates. The only real difference is that we are able to check whether a halo dof is also an interior dof.

### DIFF
--- a/source/dealii/dealii_hierarchy_helpers.cc
+++ b/source/dealii/dealii_hierarchy_helpers.cc
@@ -152,17 +152,11 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
         amge.build_agglomerate_triangulation(halo_agglomerate,
                                              halo_agglomerate_triangulation,
                                              halo_patch_to_global_map);
-        if (interior_patch_to_global_map.empty())
-        {
-          ASSERT(halo_patch_to_global_map.empty(),
-                 "If there is nothing to work on in the interior, "
-                 "the halo should also be empty.");
-          return;
-        }
+        ASSERT(interior_patch_to_global.empty() ==
+               halo_patch_to_global_map.empty());
 
         // Now that we have the triangulation, we can do the evaluation on
-        // the agglomerate
-
+        // the agglomerates
         dealii::DoFHandler<dim> interior_agglomerate_dof_handler(
             interior_agglomerate_triangulation);
         dealii::AffineConstraints<double> interior_agglomerate_constraints;
@@ -200,7 +194,7 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
             agglomerate_it - combined_agglomerate_range.begin();
         for (unsigned int j = 0; j < n_local_eigenvectors; ++j)
         {
-          unsigned int row = i * n_local_eigenvectors + j;
+          unsigned int const row = i * n_local_eigenvectors + j;
           // Get the vector used for the matrix-vector multiplication
           dealii::Vector<ScalarType> interior_delta_eig(n_interior_elem);
           for (unsigned int k = 0; k < n_interior_elem; ++k)
--- a/source/dealii/dealii_hierarchy_helpers.cc
+++ b/source/dealii/dealii_hierarchy_helpers.cc
@@ -24,6 +24,7 @@
 #include <Epetra_Map.h>
 
 #include <boost/container_hash/hash.hpp>
+#include <boost/range/combine.hpp>
 #include <boost/smart_ptr/make_unique.hpp>
 
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -126,77 +130,122 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
+        dealii::Triangulation<dim> halo_agglomerate_triangulation;
+        std::map<typename dealii::Triangulation<dim>::active_cell_iterator,
+                 typename dealii::DoFHandler<dim>::active_cell_iterator>
+            halo_patch_to_global_map;
+        amge.build_agglomerate_triangulation(halo_agglomerate,
+                                             halo_agglomerate_triangulation,
+                                             halo_patch_to_global_map);
+        if (interior_patch_to_global_map.empty())
+        {
+          ASSERT(halo_patch_to_global_map.empty(),
```
Rombur: Just use `ASSERT(interior_patch_to_global.empty() == halo_patch_to_global_map.empty())`

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -126,77 +130,122 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
+                 "If there is nothing to work on in the interior, "
+                 "the halo should also be empty.");
+          return;
+        }
+
+        // Now that we have the triangulation, we can do the evaluation on
+        // the agglomerate
+
```
Rombur: Extra blank line

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -126,77 +130,122 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
+        std::vector<dealii::types::global_dof_index> halo_dof_indices_map =
+            amge.compute_dof_index_map(halo_patch_to_global_map,
+                                       halo_agglomerate_dof_handler);
+        unsigned int const n_halo_elem = halo_dof_indices_map.size();
+
+        unsigned int const i =
+            agglomerate_it - combined_agglomerate_range.begin();
+        for (unsigned int j = 0; j < n_local_eigenvectors; ++j)
+        {
+          unsigned int row = i * n_local_eigenvectors + j;
```
Rombur: Why did you remove the `const`

source/dealii/dealii_matrix_free_hierarchy_helpers.cc
```diff
@@ -123,73 +129,113 @@ DealIIMatrixFreeHierarchyHelpers<dim, VectorType>::build_restrictor(
...
-            // Put the result in the matrix
-            // Compute the map between the local and the global dof indices.
-            std::vector<dealii::types::global_dof_index> dof_indices_map =
-                amge.compute_dof_index_map(patch_to_global_map,
-                                           agglomerate_dof_handler);
-            unsigned int const n_elem = dof_indices_map.size();
-            unsigned int const i = agglomerate_it - agglomerates_vector.begin();
-            for (unsigned int j = 0; j < n_local_eigenvectors; ++j)
+      auto worker = [&](const decltype(
+                            combined_agglomerate_range.begin()) &agglomerate_it,
```
Rombur: Since this is pretty much a copy/paste of what's above, I have the same comments.

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -205,8 +205,8 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
           dealii::Vector<ScalarType> interior_delta_eig(n_interior_elem);
           for (unsigned int k = 0; k < n_interior_elem; ++k)
           {
-            interior_delta_eig[k] = delta_eigenvector_matrix->operator()(
-                row, interior_dof_indices_map[k]);
+            interior_delta_eig[k] =
+                delta_eigenvector_matrix->el(row, interior_dof_indices_map[k]);
```
Rombur: I thought the whole point of this PR was to not use `el`

masterleinad: The documentation says that `operator()()` is expensive and `el()` should be preferred in general. I only want to able to use the former when debugging. Hence, I added a second commit where I switch back to `el()`.

masterleinad:
Let's just close this.

PR #172, opened 2019-05-07
Replace VectorDevice by dealii::LA::d::Vector<MemorySpace::CUDA>

Rombur:
This patch touches a lot of lines but it just replaces our own custom vector by deal.II's vector. Our vector was just a temporary place holder and we need to switch to deal.II's vector for matrix-free.


Rombur:
No, I guess it would have save a lot of changes :sweat_smile:

Rombur:
Can we merge this?

PR #179, opened 2019-05-15
Avoid incrementing cell iterators to often

masterleinad:
This is another factor 2 for matrix-based runs.


masterleinad:
We still spend only 3.91% in `dealii::GridTools::build_triangulation_from_patch` but 26.7% in total in `mfmg::AMGe::build_agglomerate_triangulation`. `dealii::TriaActiveIterator::operator++` takes another 13.75% and `dealii::DoFHandler::begin_active` 7.23%.

Rombur (approved):

PR #182, opened 2019-05-24
Improve memory consumption

masterleinad:
This is the run time profile I see with the changes:
```
Time before setup_restrictor: 0.301192 s
Time before delta_correction: 8.06771 s
...
Time before using agglomerates: 8.09335 s
Time before filling delta_correction: 11.8859 s
Time after compressing: 12.0269 s
Time after rescaling: 23.0697 s
...
Converging after 52 iterations.
```
and these are the old ones:
```
Time before setup_restrictor: 0.325443 s
Time before delta_correction: 8.5101 s
...
Time before using agglomerates: 8.53592 s
Time before filling delta_correction: 13.6987 s
Time after compressing: 14.6252 s
Time after rescaling: 17.4326 s
...
Converging after 52 iterations.
```
The corresponding `callgrind` shows less instructions with the changes, though.
New:
memprofile_new
Old:
memprofile_old

So we are using 70MB (~30%) less.

The part that is slower after these changes is:
```c++
    auto const range_start = eigenvector_matrix->local_range().first;
    auto const range_end = eigenvector_matrix->local_range().second;
    for (unsigned int row = range_start; row < range_end; ++row)
    {
      auto const end_iterator = eigenvector_matrix->end(row);
      for (auto column_iterator = eigenvector_matrix->begin(row);
           column_iterator != end_iterator; ++column_iterator)
      {
        column_iterator->value() *= eigenvalues[row - range_start];
      }
    }
```
which looks totally unrelated to the changes made. The memory profile indicates that we allocate and deallocate a lot here. This would at least be consistent with the `valgrind` output being better with the changes.

### DIFF
--- a/include/mfmg/common/amge.templates.hpp
+++ b/include/mfmg/common/amge.templates.hpp
@@ -179,13 +179,16 @@ void AMGe<dim, VectorType>::build_agglomerate_triangulation(
   std::vector<typename dealii::DoFHandler<dim>::active_cell_iterator>
       agglomerate;
   auto cell = _dof_handler.begin_active();
-  std::advance(cell, cell_index[0]);
-  agglomerate.push_back(cell);
-
-  for (unsigned int i = 1; i < cell_index.size(); ++i)
+  if (cell_index.size() > 0)
   {
-    std::advance(cell, cell_index[i] - cell_index[i - 1]);
+    std::advance(cell, cell_index[0]);
     agglomerate.push_back(cell);
+
+    for (unsigned int i = 1; i < cell_index.size(); ++i)
+    {
+      std::advance(cell, cell_index[i] - cell_index[i - 1]);
+      agglomerate.push_back(cell);
+    }
   }
 
   build_agglomerate_triangulation(agglomerate, agglomerate_triangulation,
--- a/include/mfmg/common/amge.templates.hpp
+++ b/include/mfmg/common/amge.templates.hpp
@@ -178,9 +178,10 @@ void AMGe<dim, VectorType>::build_agglomerate_triangulation(
 {
   std::vector<typename dealii::DoFHandler<dim>::active_cell_iterator>
       agglomerate;
-  auto cell = _dof_handler.begin_active();
+  agglomerate.reserve(cell_index.size());
   if (cell_index.size() > 0)
   {
+    auto cell = _dof_handler.begin_active();
     std::advance(cell, cell_index[0]);
     agglomerate.push_back(cell);
 
--- a/source/dealii/dealii_hierarchy_helpers.cc
+++ b/source/dealii/dealii_hierarchy_helpers.cc
@@ -237,7 +237,7 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
     // Scale the eigenvectors by their corresponding eigenvalues.
     // It turned out that accessing the individual matrix entries through the
     // deal.II TrilinosWrapper::SparseMatrix::begin() interface is much slower
-    // than constucting a representation of the eigenvalues as
+    // than constructing a representation of the eigenvalues as
     // Epetra_MultiVector and calling a Epetra_CrsMatrix member function.
     auto const range_start = eigenvector_matrix->local_range().first;
     auto const range_end = eigenvector_matrix->local_range().second;
--- a/source/dealii/dealii_hierarchy_helpers.cc
+++ b/source/dealii/dealii_hierarchy_helpers.cc
@@ -234,6 +234,11 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
     // Fill delta_correction_matrix
     delta_correction_matrix.compress(dealii::VectorOperation::insert);
 
+    // Scale the eigenvectors by their corresponding eigenvalues.
+    // It turned out that accessing the individual matrix entries through the
+    // deal.II TrilinosWrapper::SparseMatrix::begin() interface is much slower
+    // than constucting a representation of the eigenvalues as
+    // Epetra_MultiVector and calling a Epetra_CrsMatrix member function.
     auto const range_start = eigenvector_matrix->local_range().first;
     auto const range_end = eigenvector_matrix->local_range().second;
 
@@ -241,7 +246,8 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
     locally_owned_elements.add_range(range_start, range_end);
     dealii::TrilinosWrappers::MPI::Vector vector_eigenvalues(
         locally_owned_elements, comm);
-    std::vector<unsigned int> local_indices(eigenvalues.size());
+    std::vector<dealii::types::global_dof_index> local_indices(
+        eigenvalues.size());
     std::iota(local_indices.begin(), local_indices.end(), range_start);
     vector_eigenvalues.set(local_indices, eigenvalues);
     vector_eigenvalues.compress(dealii::VectorOperation::insert);
--- a/source/dealii/dealii_hierarchy_helpers.cc
+++ b/source/dealii/dealii_hierarchy_helpers.cc
@@ -122,8 +122,8 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
       } scratch_data;
       struct CopyData
       {
-        std::vector<unsigned int> rows;
-        std::vector<unsigned int> cols;
+        std::vector<dealii::types::global_dof_index> rows;
+        std::vector<dealii::types::global_dof_index> cols;
         std::vector<std::vector<dealii::TrilinosScalar>> values_per_row;
       } copy_data;
 
--- a/source/dealii/dealii_hierarchy_helpers.cc
+++ b/source/dealii/dealii_hierarchy_helpers.cc
@@ -294,14 +294,6 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
 }
 
 template <int dim, typename VectorType>
-std::shared_ptr<Operator<VectorType>>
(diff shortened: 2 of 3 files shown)


Rombur:
> The memory profile indicates that we allocate and deallocate a lot here.

What do you mean by "here"? The piece of code that you are showing? I don't see any memory allocation.

masterleinad:
This should work now. We spend around 100MB (~27%) less and run 9s (~25%) faster.
This is a comparison between the unmodified version (black), a version just exchanging the data storage type (blue), and the current version also doing the multiplication with the eigenvalues differently (red):
compare_memory

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -159,11 +157,20 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
+
+            local_copy_data.cols = amge.compute_dof_index_map(
+                patch_to_global_map, agglomerate_dof_handler);
+            auto const &dof_indices_map = local_copy_data.cols;
             unsigned int const n_elem = dof_indices_map.size();
+
+            local_copy_data.values_per_row.resize(n_local_eigenvectors);
+            std::fill(local_copy_data.values_per_row.begin(),
+                      local_copy_data.values_per_row.end(),
+                      std::vector<dealii::TrilinosScalar>(n_elem));
```
Rombur: Could this 4 lines replaces by:
```c++
local_copy_data.values_per_row.resize(n_local_eigenvectors, std::vector<dealii::TrilinosScalar>(n_elem));
```
or do you need a clean reset?

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -223,13 +232,28 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
+    std::vector<unsigned int> local_indices(eigenvalues.size());
+    std::iota(local_indices.begin(), local_indices.end(), range_start);
+    vector_eigenvalues.set(local_indices, eigenvalues);
+    vector_eigenvalues.compress(dealii::VectorOperation::insert);
+
+    Epetra_MultiVector &dummy_multi_vector =
+        vector_eigenvalues.trilinos_vector();
+    auto &dummy_matrix =
+        const_cast<Epetra_CrsMatrix &>(eigenvector_matrix->trilinos_matrix());
+    dummy_matrix.LeftScale(*dummy_multi_vector(0));
```
Rombur: This is good but we need to add a comment building the matrix this way speeds up the computation so that we don't rewrite it later to simplify the code.

source/dealii/dealii_hierarchy_helpers.cc
masterleinad: I need a clean reset. Otherwise, we would accumulate values across patches corresponding to different degrees of freedom.

include/mfmg/common/amge.templates.hpp
```diff
@@ -179,13 +179,16 @@ void AMGe<dim, VectorType>::build_agglomerate_triangulation(
...
   auto cell = _dof_handler.begin_active();
-  std::advance(cell, cell_index[0]);
-  agglomerate.push_back(cell);
-
-  for (unsigned int i = 1; i < cell_index.size(); ++i)
+  if (cell_index.size() > 0)
   {
-    std::advance(cell, cell_index[i] - cell_index[i - 1]);
+    std::advance(cell, cell_index[0]);
     agglomerate.push_back(cell);
```
aprokop: From this diff it seems that `agglomerate` is always going to be of size `cell_index.size()`. Why not resize it first and avoid `push_back`?

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -223,13 +232,34 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
-    {
-      delta_correction_matrix.set(entry.first.first, entry.first.second,
-                                  entry.second);
-    }
     delta_correction_matrix.compress(dealii::VectorOperation::insert);
 
+    // Scale the eigenvectors by their corresponding eigenvalues.
+    // It turned out that accessing the individual matrix entries through the
+    // deal.II TrilinosWrapper::SparseMatrix::begin() interface is much slower
+    // than constucting a representation of the eigenvalues as
```
aprokop: constructing

include/mfmg/common/amge.templates.hpp
masterleinad: If we are concerned about performance here, let's use `reserve` instead.

dalg24: Just reserve enough space.

aprokop: `reserve` + `push_back` is still slower than `[]`... just saying

include/mfmg/common/amge.templates.hpp
```diff
@@ -179,13 +179,16 @@ void AMGe<dim, VectorType>::build_agglomerate_triangulation(
   std::vector<typename dealii::DoFHandler<dim>::active_cell_iterator>
       agglomerate;
   auto cell = _dof_handler.begin_active();
-  std::advance(cell, cell_index[0]);
-  agglomerate.push_back(cell);
-
-  for (unsigned int i = 1; i < cell_index.size(); ++i)
+  if (cell_index.size() > 0)
```
dalg24: Did you actually run into that bug?  Should this be an assertion?

source/dealii/dealii_hierarchy_helpers.cc
dalg24: Please add a comment to capture this.

include/mfmg/common/amge.templates.hpp
dalg24: Because of checking whether new size exceed capacity?

include/mfmg/common/amge.templates.hpp
masterleinad: Yes, I ran into that. We might call this function for empty agglomerates when we use multiple `MPI` processes. I didn't look into the control flow for a way of avoiding calling the function in that case.

Rombur: It's not really a bug. The problem is that p4est can do weird things for example, if you have 4 cells you will only use one processor. The other processors won't get any cells.

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -195,23 +202,25 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
-             local_copy_data.delta_correction_local_acc)
+        for (unsigned int i = 0; i < local_copy_data.rows.size(); ++i)
         {
-          delta_correction_acc[local_pair.first] += local_pair.second;
+          // Since we didn't provide a sparsity pattern this "set" operation
+          // actually acts as an "add" operation until "compress" is called.
+          delta_correction_matrix.set(
+              local_copy_data.rows[i], local_copy_data.cols.size(),
+              local_copy_data.cols.data(),
+              local_copy_data.values_per_row[i].data(), true);
```
dalg24: Just out of curiosity, were the zero entries a significant part of the matrix?

masterleinad: I didn't check so far.

source/dealii/dealii_hierarchy_helpers.cc
```diff
@@ -223,13 +232,34 @@ DealIIHierarchyHelpers<dim, VectorType>::build_restrictor(
...
+        eigenvalues.size());
+    std::iota(local_indices.begin(), local_indices.end(), range_start);
+    vector_eigenvalues.set(local_indices, eigenvalues);
+    vector_eigenvalues.compress(dealii::VectorOperation::insert);
+
+    Epetra_MultiVector &dummy_multi_vector =
+        vector_eigenvalues.trilinos_vector();
+    auto &dummy_matrix =
+        const_cast<Epetra_CrsMatrix &>(eigenvector_matrix->trilinos_matrix());
+    dummy_matrix.LeftScale(*dummy_multi_vector(0));
```
dalg24: It feels sketchy to cast the const away.  Do you know why deal.II chose to return a const reference?

masterleinad: `const_cast` is always awful. :slightly_smiling_face:

We could modify the object in a way `deal.II` can not keep track of. That's why access to a lot of wrapped object is `const`.

include/mfmg/common/amge.templates.hpp
aprokop: push_back:
```c++
      void                                                                                                                                                                  
      push_back(const value_type& __x)                                                                                                                                      
      {                                                                                                                                                                     
    if (this->_M_impl._M_finish != this->_M_impl._M_end_of_storage)                                                                                                         
      {                                                                                                                                                                     
        _Alloc_traits::construct(this->_M_impl, this->_M_impl._M_finish,                                                                                                    
                     __x);                                                                                                                                                  
        ++this->_M_impl._M_finish;                                                                                                                                          
      }                                                                                                                                                                     
    else                                                                                                                                                                    
      _M_realloc_insert(end(), __x);                                                                                                                                        
      }   
```

operator[]:
```c++      
      reference                                                                                                                                                             
      operator[](size_type __n) _GLIBCXX_NOEXCEPT                                                                                                                           
      {                                                                                                                                                                     
    __glibcxx_requires_subscript(__n);                                                                                                                                      
    return *(this->_M_impl._M_start + __n);                                                                                                                                 
      }  

```

masterleinad: It's hard to say if initializing the elements twice is worth the faster access or not...

masterleinad:
@dalg24 told me to merge.

PR #183, opened 2019-05-28
Test multithreading

masterleinad:
This pull request makes the tester run all tests, that currently work multithreaded, with 3 threads.
We should eventually get rid off all the
```
dealii::MultithreadInfo::set_thread_limit(1); 
```


Rombur (approved):

dalg24:
Need to create an issue to keep track of these

PR #184, opened 2019-05-28
Provide virtual destructors to all polymorphic classes

masterleinad:
This became an issue when trying to debug the multithread support in one place, so I decided to be more explicitly about `virtual` destructors everywhere.


dalg24:
Why did you declare the destructors in the derived classes?

masterleinad:
Mostly to be sure that the base class destructor is `virtual`. I agree that it is not necessary to have overridden defaulted destructors, but it should also not harm.

Do you feel strongly about this?

dalg24:
> Do you feel strongly about this?

No

Issue #185, opened 2019-05-29
Make all tests work in multithreaded context


dalg24:
Need to create an issue to keep track of these

_Originally posted by @dalg24 in https://github.com/ORNL-CEES/mfmg/pull/183#issuecomment-496780222_

PR #188, opened 2019-05-30
Improve multithread support

masterleinad:
Part of #185. This pull request allows `hierarchy_driver` to run multithreaded. The main modification is that we have to copy the `Evaluator` for each agglomerate to avoid race conditions.

I have no idea why the formatting in `include/mfmg/dealii/dealii_mesh_evaluator.hpp` changed that much.

### DIFF
--- a/include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp
+++ b/include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp
@@ -57,6 +57,16 @@ public:
   virtual ~DealIIMatrixFreeMeshEvaluator() override = default;
 
   /**
+   * Create a deep copy of this class such that initializing on another
+   * agglomerate works.
+   */
+  virtual std::unique_ptr<DealIIMatrixFreeMeshEvaluator> clone() const
+  {
+    ASSERT_THROW_NOT_IMPLEMENTED();
+    return std::make_unique<DealIIMatrixFreeMeshEvaluator>(*this);
+  }
+
+  /**
    * Return the class name as std::string.
    */
   std::string get_mesh_evaluator_type() const override final;
--- a/include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp
+++ b/include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp
@@ -58,7 +58,11 @@ public:
 
   /**
    * Create a deep copy of this class such that initializing on another
-   * agglomerate works.
+   * agglomerate works. This was introduced because calls to member functions
+   * matrix_free_initialize_agglomerate() and matrix_free_evaluate_agglomerate()
+   * (implemented in user provided classes deriving from
+   * DealIIMatrixFreeMeshEvaluator) are not thread-safe. There might be other
+   * options to solve this problem.
    */
   virtual std::unique_ptr<DealIIMatrixFreeMeshEvaluator> clone() const
   {
--- a/source/dealii/dealii_matrix_free_hierarchy_helpers.cc
+++ b/source/dealii/dealii_matrix_free_hierarchy_helpers.cc
@@ -12,6 +12,7 @@
 #include <mfmg/common/instantiation.hpp>
 #include <mfmg/common/operator.hpp>
 #include <mfmg/dealii/amge_host.hpp>
+#include <mfmg/dealii/amge_host.templates.hpp>
 #include <mfmg/dealii/dealii_matrix_free_hierarchy_helpers.hpp>
 #include <mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp>
 #include <mfmg/dealii/dealii_matrix_free_operator.hpp>
@@ -83,6 +84,7 @@ DealIIMatrixFreeHierarchyHelpers<dim, VectorType>::build_restrictor(
     std::unique_ptr<dealii::TrilinosWrappers::SparseMatrix> eigenvector_matrix;
     std::unique_ptr<dealii::TrilinosWrappers::SparseMatrix>
         delta_eigenvector_matrix;
+
     amge.setup_restrictor(agglomerate_params, n_eigenvectors, tolerance,
                           *dealii_mesh_evaluator, locally_relevant_global_diag,
                           restrictor_matrix, eigenvector_matrix,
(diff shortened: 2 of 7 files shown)


tests/test_hierarchy.cc
```diff
@@ -203,7 +203,8 @@ BOOST_AUTO_TEST_CASE(benchmark)
 
 BOOST_AUTO_TEST_CASE(benchmark_mf)
 {
-  dealii::MultithreadInfo::set_thread_limit(1);
+  dealii::MultithreadInfo::set_thread_limit(
+      dealii::numbers::invalid_unsigned_int);
```
dalg24: I am getting confused on how we control multithreading.  Would you enlighten us on the precedence rule in deal.II, `DEAL_II_NUM_THREADS ` environment variable VS 3rd argument of `Utilities::MPI::MPI_InitFinalize` VS `MultithreadInfo::set_thread_limit()`, etc.

source/dealii/dealii_matrix_free_hierarchy_helpers.cc
```diff
@@ -125,91 +127,91 @@ DealIIMatrixFreeHierarchyHelpers<dim, VectorType>::build_restrictor(
...
+
+          // Perform the matrix-vector multiplication
+          dealii::Vector<ScalarType> correction(n_elem);
+          dealii::AffineConstraints<double> agglomerate_constraints;
+          using AgglomerateOperator =
+              MatrixFreeAgglomerateOperator<DealIIMatrixFreeMeshEvaluator<dim>>;
+          AgglomerateOperator agglomerate_operator(*dealii_mesh_evaluator,
+                                                   agglomerate_dof_handler,
+                                                   agglomerate_constraints);
+          agglomerate_operator.vmult(correction, delta_eig);
```
dalg24: These changes look reasonable but are lost in all the code formatting mess.

include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp
```diff
@@ -56,6 +56,16 @@ class DealIIMatrixFreeMeshEvaluator : public DealIIMeshEvaluator<dim>
...
 
+  /**
+   * Create a deep copy of this class such that initializing on another
+   * agglomerate works.
+   */
+  virtual std::unique_ptr<DealIIMatrixFreeMeshEvaluator> clone() const
+  {
+    ASSERT_THROW_NOT_IMPLEMENTED();
+    return std::make_unique<DealIIMatrixFreeMeshEvaluator>(*this);
+  }
```
dalg24: Please remind me why this cannot be pure virtual.

dalg24: Need to discuss why this is OK.

My objections are: we force the user to implement to boiler plate code, we have no way of knowing how much data he stuffed into his derived class and this could be costly.
Also I would like to understand where the race condition happen :/

tests/hierarchy_driver.cc
```diff
@@ -174,8 +174,7 @@ int main(int argc, char *argv[])
 {
   namespace boost_po = boost::program_options;
 
-  MPI_Init(&argc, &argv);
-  dealii::MultithreadInfo::set_thread_limit(1);
+  dealii::Utilities::MPI::MPI_InitFinalize mpi_init(argc, argv);
```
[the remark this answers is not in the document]
masterleinad: For requesting explicit thread support we need to call `MPI_Init_thread` instead which `dealii::Utilities::MPI::MPI_InitFinalize` does (apart from initializing Zoltan and emptying memory pools).

`MPI_Init` seems to not allow calling `MPI` functions from multiple threads and hence behaves similar to `MPI_Init_thread` with `MPI_THREAD_SINGLE` (or possibly `MPI_THREAD_FUNNELED`).

In the end, I was just seeing `MPI` errors when calling `MPI` functions (inside the `MatrixFree` constructior to be precise) when running multithreaded. Using `MPI_Init` instead of `MPI_Init_thread` or `dealii::Utilities::MPI::MPI_InitFinalize` was likely just an oversight.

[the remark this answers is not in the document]
dalg24: Updated link to deal.II code

tests/test_hierarchy.cc
masterleinad: `dealii::MultithreadInfo::set_thread_limit` sets the number of threads to the minimum its the argument and the environment variable. `TBB` is asked if the minimum is `dealii::numbers::invalid_unsigned_int`.
See [here](https://www.dealii.org/current/doxygen/deal.II/multithread__info_8cc_source.html#l00108). That function is called before `main` is executed.
 `dealii::Utilities::MPI::MPI_InitFinalize` just calls this function using its third argument (`dealii::numbers::invalid_unsigned_int` by default).

Consequently, the call here sets the number of threads to the one stored in the environment variable.

include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp
masterleinad: This would be much easier if we have a uniform interface for global/agglomerate initialization and evaluation or separate user classes.

masterleinad:
```
/usr/include/c++/7/ext/new_allocator.h:136:4: error: invalid new-expression of abstract class type ‘TestMeshEvaluator<mfmg::DealIIMatrixFreeMeshEvaluator<2> >’
  { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from mfmg/tests/test_hierarchy.cc:37:0:
mfmg/tests/test_hierarchy_helpers.hpp:202:7: note:   because the following virtual functions are pure within ‘TestMeshEvaluator<mfmg::DealIIMatrixFreeMeshEvaluator<2> >’:
 class TestMeshEvaluator final : public MeshEvaluator
       ^~~~~~~~~~~~~~~~~
In file included from mfmg/include/mfmg/dealii/amge_host.hpp:16:0,
                 from mfmg/include/mfmg/dealii/dealii_hierarchy_helpers.hpp:16,
                 from mfmg/include/mfmg/common/hierarchy.hpp:19,
                 from mfmg/tests/test_hierarchy.cc:14:
mfmg/include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp:63:58: note: 	std::unique_ptr<mfmg::DealIIMatrixFreeMeshEvaluator<dim> > mfmg::DealIIMatrixFreeMeshEvaluator<dim>::clone() const [with int dim = 2]
   virtual std::unique_ptr<DealIIMatrixFreeMeshEvaluator> clone() const = 0;
                                                          ^~~~~
```
I am not sure if we need to modify the matrix-based version as well.

tests/test_hierarchy_helpers.hpp
```diff
@@ -309,8 +309,23 @@ class TestMFMeshEvaluator final
...
   }
 
+  TestMFMeshEvaluator(
+      TestMFMeshEvaluator<dim, fe_degree, ScalarType> const &_other_evaluator)
+      : mfmg::DealIIMatrixFreeMeshEvaluator<dim>(*this),
+        _material_property(_other_evaluator._material_property),
+        _fe(_other_evaluator._fe),
+        _laplace_operator(_other_evaluator._laplace_operator)
+  {
+  }
```
dalg24: Why did you define the copy constructor?

masterleinad:
The matrix-based version needs more work, so I restricted the number of threads to one in that case again, To actually check the matrix-free version here, I changed the default as well to "matrix-free" as well (forcing a Chebyshev smoother since that is the only possibility).

tests/test_hierarchy_helpers.hpp
masterleinad: We call the copy constructor in `clone()` and I wanted to make sure we copy as little as possible here. In particular, we should need to copy `mutable` members since they are set up in `matrix_free_initialize_agglomerate` only.

dalg24: Please add a comment

Rombur:
So the problem is that `MatrixFree` is not thread-safe? I don't get why we need to clone the `MeshEvaluator`

masterleinad:
> So the problem is that `MatrixFree` is not thread-safe? I don't get why we need to clone the `MeshEvaluator`

No, that is not the problem. So far, we had a common `Evaluator` object that is both responsible for agglomerate evaluations as well as global evaluations. If we use multiple threads, we try to evaluate different agglomerates at the same time. In particular, we change the state of the class in `initialize_agglomerate` which is then used in `evaluate_agglomerate`. Hence, all the agglomerate related member variables might be involved in race conditions. Apart from that, we might run into an inconsistent state comparing `initialize_agglomerate` and `evaluate_agglomerate`.

masterleinad:
As compared to https://github.com/ORNL-CEES/mfmg/pull/192 changes in timings for running with one thread are unaffected. These are the run times on my notebook:

| # threads | 2x2 | 4x4 |
|--------------|------|------| 
|1 | 37.3 |20.5|
|2 | 25.8| 12.6|
|4| 13.4| 8.9|
|8|11.9| 8.32|

masterleinad:
For the 4x4 patch and one thread, I get

| Section                         | no. calls |  wall time | % of total |
|--|--|--|--|
| Apply                           |        74 |      2.01s |        10% |
| Apply: fine levels              |        74 |         2s |        10% |
| Setup                           |         1 |      14.8s |        75% |
| Setup: build restrictor         |         1 |      14.8s |        75% |
 
while 8 threads give me

| Section                         | no. calls |  wall time | % of total |
|--|--|--|--|
| Apply                           |        74 |      1.92s |        23% |
| Apply: fine levels              |        74 |      1.91s |        23% |
| Setup                           |         1 |      3.96s |        48% |
| Setup: build restrictor         |         1 |      3.95s |        48% |

masterleinad:
Using vectorization on top of that gives me (1 thread)
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      7.73s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Apply                           |        74 |     0.557s |       7.2% |
| Apply: fine levels              |        74 |     0.553s |       7.2% |
| Setup                           |         1 |      6.88s |        89% |
| Setup: build restrictor         |         1 |      6.88s |        89% |
+---------------------------------+-----------+------------+------------+

resp. (8 threads)

+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      2.73s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Apply                           |        74 |       0.6s |        22% |
| Apply: fine levels              |        74 |     0.596s |        22% |
| Setup                           |         1 |      1.84s |        67% |
| Setup: build restrictor         |         1 |      1.83s |        67% |
+---------------------------------+-----------+------------+------------+

**Comment by masterleinad** [comment]
Using MPI instead of threads gives (1 MPI process)
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      8.89s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Apply                           |        74 |     0.665s |       7.5% |
| Apply: fine levels              |        74 |     0.659s |       7.4% |
| Setup                           |         1 |      7.88s |        89% |
| Setup: build restrictor         |         1 |      7.87s |        89% |
+---------------------------------+-----------+------------+------------+
```
resp. (8 MPI processes)
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      2.12s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Apply                           |        74 |     0.349s |        16% |
| Apply: fine levels              |        74 |     0.331s |        16% |
| Setup                           |         1 |      1.62s |        76% |
| Setup: build restrictor         |         1 |      1.61s |        76% |
+---------------------------------+-----------+------------+------------+
```

**Comment by masterleinad** [comment]
Does anyone want to discuss a different solution?

**Comment by Rombur** [comment]
@dalg24 said he was going to look at it

File: source/dealii/dealii_matrix_free_hierarchy_helpers.cc [hunk 6]
```diff
@@ -12,6 +12,7 @@
 #include <mfmg/common/instantiation.hpp>
 #include <mfmg/common/operator.hpp>
 #include <mfmg/dealii/amge_host.hpp>
+#include <mfmg/dealii/amge_host.templates.hpp>
```
dalg24: Comment it is included for `MatrixFreeAgglomerateOperator` and that it would probably not be a bad idea to move the definition elsewhere.

include/mfmg/dealii/dealii_matrix_free_mesh_evaluator.hpp
```diff
@@ -56,6 +56,16 @@ class DealIIMatrixFreeMeshEvaluator : public DealIIMeshEvaluator<dim>
    */
   virtual ~DealIIMatrixFreeMeshEvaluator() override = default;
 
+  /**
+   * Create a deep copy of this class such that initializing on another
+   * agglomerate works.
```
dalg24: Please add a note that this was introduced because calls to member functions `matrix_free_initialize_agglomerate()` and `matrix_free_evaluate_agglomerate()` (implemented in user provided class deriving from `MatrixFreeMeshEvaluator`) are not thread safe and that it is not the only option to solve the problem.

masterleinad:
Thanks!

PR #190, opened 2019-05-30
Allow for regular Triangulations

masterleinad:
With this change, we can also use regular `Triangulation` objects.

### DIFF
--- a/source/dealii/dealii_matrix_free_mesh_evaluator.cc
+++ b/source/dealii/dealii_matrix_free_mesh_evaluator.cc
@@ -36,10 +36,13 @@ DealIIMatrixFreeMeshEvaluator<dim>::build_range_vector() const
   auto const &triangulation = (this->_dof_handler).get_triangulation();
   using Triangulation =
       typename std::remove_reference<decltype(triangulation)>::type;
-  auto comm = static_cast<dealii::parallel::Triangulation<
-      Triangulation::dimension, Triangulation::space_dimension> const &>(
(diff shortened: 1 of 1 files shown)


source/dealii/dealii_matrix_free_mesh_evaluator.cc
```diff
@@ -36,10 +36,13 @@ DealIIMatrixFreeMeshEvaluator<dim>::build_range_vector() const
   auto const &triangulation = (this->_dof_handler).get_triangulation();
   using Triangulation =
       typename std::remove_reference<decltype(triangulation)>::type;
-  auto comm = static_cast<dealii::parallel::Triangulation<
-      Triangulation::dimension, Triangulation::space_dimension> const &>(
-                  triangulation)
-                  .get_communicator();
+  MPI_Comm comm;
+  if (const auto parallel_tria = dynamic_cast<dealii::parallel::Triangulation<
```
dalg24: West `const` or no `const` (my preference goes to the latter)

masterleinad:
I thought it would be a nice addition to go in a direction where we don't strictly enforce `MPI`. In particular, this would make debugging easier.

Feel free to close if you don't agree.

Rombur:
> I thought it would be a nice addition to go in a direction where we don't strictly enforce MPI.

We decided at the beginning of the project that MPI will always be required. Given the number of libraries we depend on. Getting rid of MPI won't make anything easier.

PR #192, opened 2019-05-30
Allow using SolverCG in hierarchy_driver

masterleinad:
This pull request makes `hierarchy_driver` look closer to what we are using for profiling.
Essentially, we just test for `Hierarchy` as a preconditioner for `CG` instead of as a solver on its own.

Apart from that also some of the default values are changed. We use polynomial degree 4 rather than one, and enable `Lanczos` as eigensolver and use the `FastAP` interface.

### DIFF
--- a/tests/hierarchy_driver.cc
+++ b/tests/hierarchy_driver.cc
@@ -225,6 +225,8 @@ int main(int argc, char *argv[])
   cmd.add_options()("filename,f", boost_po::value<std::string>()->multitoken(),
                     "file containing input parameters");
   cmd.add_options()("dim,d", boost_po::value<int>(), "dimension");
+  cmd.add_options()("matrix_free,m", boost_po::value<bool>(),
+                    "use matrix-free algorithm");
   cmd.add_options()("tolerance,t", boost_po::value<double>(),
                     "tolerance to use for the solver");
 
@@ -248,6 +250,10 @@ int main(int argc, char *argv[])
     dim = vm["dim"].as<int>();
   mfmg::ASSERT(dim == 2 || dim == 3, "Dimension must be 2 or 3");
 
+  bool matrix_free = false;
+  if (vm.count("matrix_free"))
+    matrix_free = vm["matrix_free"].as<bool>();
+
   auto const params = std::make_shared<boost::property_tree::ptree>();
   boost::property_tree::info_parser::read_info(filename, *params);
 
@@ -262,15 +268,13 @@ int main(int argc, char *argv[])
   }
   params->put("solver.tolerance", solver_tolerance);
 
-  // matrix-free
-  for (bool preconditioner : {true, false})
+  std::cout << "input file: " << filename << ", dimension: " << dim
+            << ", matrix-free: " << matrix_free << ", fe_degree: " << fe_degree
+            << ", solver_tolerance: " << solver_tolerance << std::endl;
+
+  if (matrix_free)
   {
-    params->put("is preconditioner", preconditioner);
     params->put("smoother.type", "Chebyshev");
-    std::cout << "input file: " << filename << ", dimension: " << dim
-              << ", matrix-free: true, preconditioner: " << preconditioner
-              << ", fe_degree: " << fe_degree
-              << ", solver_tolerance: " << solver_tolerance << std::endl;
 
     if (dim == 2)
     {
@@ -413,16 +417,8 @@ int main(int argc, char *argv[])
       }
     }
   }
-
-  // matrix-based
-  for (bool preconditioner : {true, false})
+  else
   {
-    params->put("is preconditioner", preconditioner);
-    params->put("smoother.type", "Gauss-Seidel");
-    std::cout << "input file: " << filename << ", dimension: " << dim
-              << ", matrix-free: false, preconditioner: " << preconditioner
-              << ", fe_degree: " << fe_degree
-              << ", solver_tolerance: " << solver_tolerance << std::endl;
     if (dim == 2)
       matrix_based_two_grids<2>(params);
     else
--- a/tests/hierarchy_driver.cc
+++ b/tests/hierarchy_driver.cc
@@ -119,6 +119,8 @@ void matrix_free_two_grids(std::shared_ptr<boost::property_tree::ptree> params)
 template <int dim>
 void matrix_based_two_grids(std::shared_ptr<boost::property_tree::ptree> params)
 {
+  bool const test_preconditioner = params->get<bool>("is preconditioner");
+
   using DVector = dealii::LinearAlgebra::distributed::Vector<double>;
   using MeshEvaluator = mfmg::DealIIMeshEvaluator<dim>;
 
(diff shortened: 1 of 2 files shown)


tests/hierarchy_driver.cc
```diff
@@ -63,33 +64,48 @@ void matrix_free_two_grids(std::shared_ptr<boost::property_tree::ptree> params)
...
+          << conv_rate << std::endl;
+  }
+  else
+  {
+    dealii::SolverControl solver_control(solution.size(), 1.e-6, true, true);
+    dealii::SolverCG<DVector> solver(solver_control);
+
+    solver.solve(mf_laplace._laplace_operator, solution, rhs,
+                 //                 dealii::PreconditionIdentity()
+                 hierarchy);
```
dalg24: Did you mean to leave that commented line?

tests/hierarchy_driver.cc
```diff
@@ -114,7 +130,7 @@ void matrix_based_two_grids(std::shared_ptr<boost::property_tree::ptree> params)
   Source<dim> source;
 
   auto laplace_ptree = params->get_child("laplace");
-  auto fe_degree = laplace_ptree.get<unsigned>("fe_degree", 1);
+  auto fe_degree = laplace_ptree.get<unsigned>("fe_degree", 4);
```
dalg24: What was the motivation for changing the default value here?

tests/hierarchy_driver.cc
```diff
@@ -63,33 +64,48 @@ void matrix_free_two_grids(std::shared_ptr<boost::property_tree::ptree> params)
...
 
-  mf_laplace._laplace_operator.vmult(residual, solution);
-  residual.sadd(-1., 1., rhs);
-  auto const residual0_norm = residual.l2_norm();
+  mfmg::Hierarchy<DVector> hierarchy(comm, evaluator, params, timer);
 
-  std::cout << std::scientific;
-  res[0] = 1.0;
-  for (unsigned int i = 0; i < n_cycles; ++i)
+  if (!test_preconditioner)
```
dalg24: I don't find `"is preconditioner"` not very descriptive.  Please comment that it let the user specify whether to solve the Laplace problem and report the number of iterations required or whether to perform a set number of V-cycles and report the convergence rate.

tests/hierarchy_driver.cc
```diff
@@ -212,7 +228,9 @@ int main(int argc, char *argv[])
   auto const params = std::make_shared<boost::property_tree::ptree>();
   boost::property_tree::info_parser::read_info(filename, *params);
 
-  int const fe_degree = params->get<unsigned int>("laplace.fe_degree", 1);
+  int const fe_degree = params->get<unsigned int>("laplace.fe_degree", 4);
+  params->put("fast_ap", true);
+  params->put("eigensolver.type", "lanczos");
```
dalg24: What was the default?

tests/hierarchy_driver.cc
masterleinad: "is preconditioner" is preexisting. Its main purpose is to decide whether to zero out the destination vector in `Hierarchy::apply`.

tests/hierarchy_driver.cc
masterleinad: `lanczos` for matrix-free and `arpack` for matrix-based. I just like to be explicit about the parameters we are using here.

tests/hierarchy_driver.cc
masterleinad: Yes, that was on purpose. I see this test as the one we want to profile with. Hence, I like to have a reminder for what I need to change if I don't want a preconditioner.

Anyway, I can remove it if you like that better.

tests/hierarchy_driver.cc
masterleinad: I just wanted to be more close to what we use for profiling, but don't feel strongly about this.

tests/hierarchy_driver.cc
```diff
@@ -20,6 +20,7 @@
 template <int dim, int fe_degree>
 void matrix_free_two_grids(std::shared_ptr<boost::property_tree::ptree> params)
 {
+  const bool test_preconditioner = params->get("is preconditioner", true);
```
Rombur: left const. I thought you were one of us...

tests/hierarchy_driver.cc
dalg24: OK to leave if you add a comment above

tests/hierarchy_driver.cc
masterleinad: Just try
```
 grep -rIn "  const " source/ include/ tests/
```

tests/hierarchy_driver.cc
masterleinad: I changed it back.

masterleinad:
I guess it would be beneficial to merge this before #188 so we can profile that one more easily.

tests/hierarchy_driver.cc
```diff
@@ -63,33 +69,49 @@ void matrix_free_two_grids(std::shared_ptr<boost::property_tree::ptree> params)
...
+      res[i + 1] = rel_residual;
+    }
+
+    double const conv_rate = res[n_cycles] / res[n_cycles - 1];
+    pcout << "Convergence rate: " << std::fixed << std::setprecision(2)
+          << conv_rate << std::endl;
+  }
+  else
+  {
+    dealii::SolverControl solver_control(solution.size(), 1.e-6, true, true);
```
aprokop: Make this tolerance a command line argument.

masterleinad: Why do we need even more command line arguments? It's really difficult to keep track of all the possible ways to provide parameters already. I would actually prefer if we hardcode the configuration for every test file instead of using `hierarchy_input.info` at all.

aprokop: Because it would be helpful to study number of iterations as a function of tolerance. It would be fine to have it as a parameter in `hierarchy_input.info`.

Rombur (approved):

masterleinad:
@aprokop I pushed a commit to make the solver tolerance a command line argument.

aprokop (approved):

aprokop (changes requested):
The low coverage on diff indicates that the code devoted to matrix-free is not being executed in the testing. Please add.

masterleinad:
You just want `matrix-free` to be the default?

aprokop:
> You just want `matrix-free` to be the default?

I don't care about the default, just that the tests need to exercise both branches.

masterleinad:
Are you referring to `matrix-free` vs `matrix-based`, or `preconditioner` vs `solver`? Or both?
We didn't test for `matrix-free` before.

aprokop:
The main point is about `matrix-free` vs `matrix-based`, as they have 20+ lines independent code paths. Testing `preconditioner` is required, as you introduce CG solver. Testing `solver` is optional as we have `test_hierarchy`.

masterleinad:
Is that what you had in mind?

aprokop (approved):
Yep, this looks good.

PR #196, opened 2019-06-03
Fix a bug where MFMG_VERSION was set after being used

### LINKED ISSUE #122: MFMG_VERSION used before defined in root CMake
```
CMake Warning at CMakeLists.txt:3 (PROJECT):
  VERSION keyword not followed by a value or was followed by a value that
  expanded to nothing.
```
The problem seems that `SetupMFMG` is being included *after* `MFMG_VERSION` is used.

Rombur:
Fix #122


dalg24:
`cmake/SetupMFMG.cmake` makes little sense.  I think `-g` will be there by default when you configure with `-D CMAKE_BUILD_TYPE=Debug` or `...=RelWithDebInfo`.  All we are left with is the logic to enable or disable the runtime assertions.

PR #198, opened 2019-06-05
Enforce disabled ANASAZI_TEUCHOS_TIME_MONITOR

masterleinad:
I just get segmentation faults when running with multiple threads if this preprocessor flag is active.


masterleinad:
When changing the corresponding `CMake` option, building didn't do anything. I am not sure if we can just undefine the variable in our code safely.

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -215,6 +215,11 @@ void anasazi_compute_eigenvalues_and_eigenvectors(
     std::vector<std::complex<double>> &eigenvalues,
     std::vector<dealii::Vector<double>> &eigenvectors)
 {
+#ifdef ANASAZI_TEUCHOS_TIME_MONITOR
```
dalg24: Would that be defined by Trilinos?

dalg24: Is it something that we could/should treat at configuration time or are we just making sure we don't accidentally call that function at runtime?

masterleinad: We should just configure `Trilinos` appropriately.

masterleinad:
```
//Build with Teuchos TimeMonitors enabled.
Anasazi_ENABLE_TEUCHOS_TIME_MONITOR:BOOL=OFF
```

cmake/SetupTPLs.cmake
```diff
@@ -19,6 +19,26 @@ IF(NOT DEAL_II_WITH_TRILINOS)
           "Error! deal.II must be compiled with Trilinos support.")
 ENDIF()
 
+INCLUDE(CheckCXXSourceCompiles)
+SET(CMAKE_REQUIRED_INCLUDES ${DEAL_II_INCLUDE_DIRS})
+CHECK_CXX_SOURCE_COMPILES(
```
dalg24: Trilinos include directories come from the deal.II variable?

masterleinad: Yes, we find `Trilinos` only through `deal.II`.

masterleinad:
The updated container is at https://hub.docker.com/r/masterleinad/mfmg-stack.

Rombur:
@masterleinad  just push it here: https://cloud.docker.com/repository/docker/rombur/mfmg-stack/general with the tag 19.06.0

masterleinad:
I pushed it to your repository as well.

cmake/SetupTPLs.cmake
```diff
@@ -19,6 +19,28 @@ IF(NOT DEAL_II_WITH_TRILINOS)
...
+  #ifdef ANASAZI_TEUCHOS_TIME_MONITOR
+  error
+  #endif
+  int main(){}
+  "
+  MFMG_APPROPRIATE_ANASAZI
+  )
+UNSET(CMAKE_REQUIRED_INCLUDES)
+IF(NOT MFMG_APPROPRIATE_ANASAZI)
+  MESSAGE(FATAL_ERROR
```
Rombur: Don't use `FATAL_ERROR`, we use `SEND_ERROR`

PR #201, opened 2019-06-12
Fix DiscontinuousMaterialProperty

### LINKED ISSUE #199: Test all material_proprety options
Currently, the solver doesn't converge for `discontinuous`, but gives similar results for `constant`, `linear` and `linear_x`. We should make sure that all options actually work.

masterleinad:
This change makes sure that we are indeed using the same coefficients for the matrix-free and the matrix-based case. Fixes #199.

### DIFF
--- a/tests/test_hierarchy.cc
+++ b/tests/test_hierarchy.cc
@@ -597,3 +597,65 @@ BOOST_AUTO_TEST_CASE(fast_multiply_transpose_mf)
         BOOST_TEST(fast_matrix->el(i, j) == ref_matrix->el(i, j),
                    tt::tolerance(1e-9));
 }
+
+template <int dim>
+class ConstantMaterialProperty;
+template <int dim>
+class LinearMaterialProperty;
+template <int dim>
+class LinearXMaterialProperty;
+template <int dim>
+class DiscontinuousMaterialProperty;
+
+using material_properties =
+    std::tuple<ConstantMaterialProperty<2>, LinearMaterialProperty<2>,
+               LinearXMaterialProperty<2>, DiscontinuousMaterialProperty<2>>;
+BOOST_AUTO_TEST_CASE_TEMPLATE(test_material_properties, MaterialProperty,
+                              material_properties)
+{
+  dealii::MultithreadInfo::set_thread_limit(1);
+
+  MPI_Comm comm = MPI_COMM_WORLD;
+
+  using DVector = dealii::LinearAlgebra::distributed::Vector<double>;
+  int constexpr dim = mfmg::DealIIMatrixFreeMeshEvaluator<2>::_dim;
+
+  auto params = std::make_shared<boost::property_tree::ptree>();
+  boost::property_tree::info_parser::read_info("hierarchy_input.info", *params);
+  auto const material_property = std::make_shared<MaterialProperty>();
+  Source<dim> source;
+  auto laplace_ptree = params->get_child("laplace");
+
+  // Set up the reference (matrix-based) operator.
+  Laplace<dim, DVector> ref_laplace(comm, 1);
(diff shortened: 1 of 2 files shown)


tests/test_hierarchy_helpers.hpp
```diff
@@ -155,11 +155,19 @@ class DiscontinuousMaterialProperty final : public Coefficient<dim>
...
+    for (unsigned int i = 0;
+         i < dealii::VectorizedArray<double>::n_array_elements; ++i)
+      for (unsigned int d = 0; d < dim; ++d)
+        dim_scale[i] +=
+            static_cast<unsigned int>(std::floor(100. * p[d][i])) % 2;
+
+    dealii::VectorizedArray<double> return_value;
+    for (unsigned int i = 0;
+         i < dealii::VectorizedArray<double>::n_array_elements; ++i)
+      return_value[i] = (dim_scale[i] == dim ? 100. : 10.);
```
Rombur: Can you assign a `double` to a `VectorizedArray`?

Rombur (commented):
It looks OK but it doesn't compile on jenkins (`one` is not used).

tests/test_hierarchy_helpers.hpp
masterleinad: Yes, https://www.dealii.org/current/doxygen/deal.II/classVectorizedArray.html#a2a6e8e58f020c57afc202b17f93c0774, but that is not what I am doing.

tests/test_hierarchy.cc
```diff
@@ -597,3 +597,65 @@ BOOST_AUTO_TEST_CASE(fast_multiply_transpose_mf)
...
 }
+
+template <int dim>
+class ConstantMaterialProperty;
+template <int dim>
+class LinearMaterialProperty;
+template <int dim>
+class LinearXMaterialProperty;
+template <int dim>
+class DiscontinuousMaterialProperty;
```
dalg24: Why did you add these declarations?  Aren't they already available in the header `test_hierarchy_helpers.hpp`?

tests/test_hierarchy.cc
```diff
@@ -597,3 +597,65 @@ BOOST_AUTO_TEST_CASE(fast_multiply_transpose_mf)
...
+               LinearXMaterialProperty<2>, DiscontinuousMaterialProperty<2>>;
+BOOST_AUTO_TEST_CASE_TEMPLATE(test_material_properties, MaterialProperty,
+                              material_properties)
+{
+  dealii::MultithreadInfo::set_thread_limit(1);
+
+  MPI_Comm comm = MPI_COMM_WORLD;
+
+  using DVector = dealii::LinearAlgebra::distributed::Vector<double>;
+  int constexpr dim = mfmg::DealIIMatrixFreeMeshEvaluator<2>::_dim;
```
dalg24: ??

tests/test_hierarchy.cc
masterleinad: Ah, yes they made debugging easier. I'll remove them.

tests/test_hierarchy.cc
masterleinad: We do that all over the place in this test.

dalg24: I see two other offenders: https://github.com/ORNL-CEES/mfmg/blob/93e4253727b92d7dbc3d4dd7c743bbaf6c70a9c9/tests/test_hierarchy.cc#L472 and https://github.com/ORNL-CEES/mfmg/blob/93e4253727b92d7dbc3d4dd7c743bbaf6c70a9c9/tests/test_hierarchy.cc#L540

Please change to `int constexpr dim = <DIM>;` and let's merge

dalg24 (approved):

PR #202, opened 2019-06-13
Compare with ML

dalg24:


include/mfmg/common/hierarchy.hpp
```diff
@@ -181,6 +183,13 @@ class Hierarchy
 
       if (level_index == num_levels - 1)
       {
+        if (level_index == 0)
+        {
+          // When using ML for the full hierarchy, do not zero out initial guess
+          params->put("coarse.params.zero starting solution",
```
dalg24: @aprokop I could not find `"zero starting solution"` in ML's documentation.  Where did you find out about it?

aprokop: I could not find anything about using ML in standalone mode in the documentation either. I had to dig in in the source code of `ml/src/Utils/ml_MultiLevelPreconditioner.{h,cpp}` to find out about it.

include/mfmg/common/hierarchy.hpp
```diff
@@ -170,6 +170,8 @@ class Hierarchy
 
     // TODO: add stopping criteria for levels (number of levels / coarse size)
     int const num_levels = params->get("max levels", 2);
+    ASSERT(num_levels > 0,
+           "\"max levels\" parameter must be greater than or equal to one");
```
aprokop:
```suggestion
           "number of levels specified by \"max levels\" parameter must be positive");
```

tests/data/hierarchy_input.info
```diff
@@ -30,3 +30,17 @@ laplace
   reordering None
   n_refinements 5
 }
+
+;; Uncomment below for running ML
```
aprokop: Why not create `hierarchy_input_ml.info` input deck that is actually tested?

tests/test_hierarchy.cc
```diff
@@ -259,6 +259,25 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(ml, MeshEvaluator, mesh_evaluator_types)
...
+  // params->put("coarse.params.max levels", 3); // control number of ML levels
+  // if wanted
+  params->put(
+      "coarse.params.coarse: type",
+      "Amesos-KLU"); // for some reason, without specifying this I'm getting
+                     // MLS/Chebyshev on coarse ML, which is not what we want
+  params->put("coarse.params.smoother: type",
+              "symmetric Gauss-Seidel"); // works a lot better than default
+                                         // Chebyshev/MLS
+  BOOST_TEST( test<mfmg::DealIIMeshEvaluator<2>>(params) < 0.1 );
```
aprokop: Have you tested this with large `n_refinements` values?

dalg24: No.  Why would I?

Rombur:
Can you do one run of the benchmark problem that you guys have using? So that we get an idea of how we compare to ML.

aprokop:
@dalg24 Do you have a follow up PR for this to add it to the hierarchy driver? I don't see it here, and I thought we discussed to put that in.

dalg24:
> @dalg24 Do you have a follow up PR for this to add it to the hierarchy driver? I don't see it here, and I thought we discussed to put that in.

Uncomment the block and you've got it

masterleinad:
```
The following tests FAILED:
	  1 - check_format (Failed)
	 11 - test_hierarchy_2 (Failed)
	 12 - test_hierarchy_4 (Failed)
```

tests/data/hierarchy_input.info
```diff
@@ -30,3 +30,17 @@ laplace
...
+;"max levels" 1
+;coarse
+;{
+;  type ML
+;  params
+;  {
+;    "ML output" 43
+;    "max levels" 3
```
Rombur: @aprokop Shouldn't this be 2 instead of 3?

dalg24:
> The following tests FAILED:
> 	  1 - check_format (Failed)
> 	 11 - test_hierarchy_2 (Failed)
> 	 12 - test_hierarchy_4 (Failed)

Fixed the format and "bumped" convergence rate from .1 to .2

tests/data/hierarchy_input.info
aprokop: This specifies the number of levels in ML hierarchy. I don't think it should be specified at all, allowing ML to determine it on its own.

dalg24: Will remove

tests/test_hierarchy.cc
dalg24: (Is not actually "outdated", I just applied clang-format)

tests/data/hierarchy_input.info
dalg24: Haven't put much thought into it.  I am not so keen on having two files that are almost the same.
That said, INFO format support include directives so I could move the ML specific parameters into a separate file and have a commented include directive here.  I don't think it makes things better though.

masterleinad:
This is approximately 4 times slower than the current matrix-free setup (vectorized+threads).

aprokop:
> This is approximately 4 times slower than the current matrix-free setup (vectorized+threads).

Well, it's serial, so should be compared to a single thread.

Rombur:
The multithreaded code is about 4 times faster than serial. So we are in the same ballpark.

masterleinad:
I should have said that I measure around 8s (compare to https://github.com/ORNL-CEES/mfmg/pull/188#issuecomment-498842991 and below).

tests/data/hierarchy_input.info
masterleinad: Uncommenting these lines breaks matrix-free support (in `hierarrchy_driver`). I don't really like that. I would appreciate an approach where we can only tell the matrix-based version to use a different path.

dalg24: Fixed

PR #203, opened 2019-06-14
Patch for avoiding reinitializing ShapeInfo

masterleinad:


dalg24:
Need to build a new image for it.

masterleinad:
> Need to build a new image for it.

On it.

masterleinad:
Done.

dalg24:
You need to update the image tag in the Jenkins configuration

tests/test_hierarchy_helpers.hpp
```diff
@@ -406,8 +406,10 @@ class TestMFMeshEvaluator final
                                            dealii::update_quadrature_points;
     std::shared_ptr<dealii::MatrixFree<dim, ScalarType>> mf_storage(
         new dealii::MatrixFree<dim, ScalarType>());
-    mf_storage->reinit(dof_handler, _agg_constraints,
-                       dealii::QGauss<1>(fe_degree + 1), additional_data);
+    mf_storage->reinit(
```
Rombur: Let's protect this code with a macro so that if deal.II is not patched we can still run the test. A simple macro defined in the file to switch the implementation is good enough.

masterleinad: Good idea.

Rombur (approved):

tests/test_hierarchy_helpers.hpp
```diff
@@ -34,6 +34,12 @@
 #include "laplace.hpp"
 #include "laplace_matrix_free.hpp"
 
+// If activated the global operator's ShapeInfo object is copied instead of
+// recalculated when initializing the agglomerate operators. Requires that
+// deal.II is patched with scripts/0002-Avoid-copying-shape-info.patch.
+//
+// #define USE_OPTIMIZED_AGGLOMERATE_INITIALIZATION
```
dalg24: How did you intent to enable the optimization?
One option would have been to define it in deal.II's header you patched but it involves rebuilding the image.
Another might be adding the compile definition using CMake.

Rombur: Just uncomment and recompile. I don't expect us to switch between the two often, i.e., performance test,  you uncomment otherwise you don't care

PR #204, opened 2019-06-14
[WIP] Matrix-free cuda

Rombur:
This PR is very much a WIP, there is no need to look at it except for @masterleinad :) The things that need to be done in order of priority:
- [x] Make this [test](https://github.com/ORNL-CEES/mfmg/compare/master...Rombur:mf_cuda?expand=1#diff-f09c6561ce5989c55afa687dd4f6bf10R274 ) work in parallel: the good news is that it works in serial the bad news is that we actually never ran that test in parallel and so even the matrix-based behaves in a strange way (i.e. only the first step decreases the error). In matrix-free, there is a segfault on rank 1 [here](https://github.com/ORNL-CEES/mfmg/blob/master/include/mfmg/common/amge.templates.hpp#L316) in `locally_relevant_global_diag`. `locally_relevant_global_diag` is computed on the host because we can't get the diagonal from matrix-free with cuda.
- [x] I have only tested in serial with 1 eigenvector, so we need to check what happens with 2 eigenvectors. There is no need to use more eigenvectors because we know that lanczos is very brittle.
- [ ] ~Right now we don't use `material_property` when using cuda matrix-free, so we need to add that.~ This can be done in another PR
- [x] I have had to make a few changes in the interface of several functions. Which means that only one test compiles, the others are commented. We need to uncomment them and make them pass again.
- [x] We need to rebase on master.

This is all the work that still needs to be done on this PR. Once it's done we can do some cleaning and split this PR in smaller ones. If you can tackle the first two points that would be fantastic but feel free to do more if you are bored :smile:


masterleinad:
AFAICT the only problem left is with the coarse solver, i.e. replacing it by the identity gives the same results for 1 and 2 MPI processes.

masterleinad:
I see the same test failure locally as `ci` reports:
```
/var/jenkins/workspace/mfmg-experimental/tests/test_hierarchy_device.cu(308): [NON-XML-CHAR-0x1B][1;31;49merror: in "amgx/_0": difference{3.16667} between conv_rate{1} and ref_solution{0.23999999999999999} exceeds 5%
Failure occurred in a following context:
    mesh_evaluator_type = matrix_based; [NON-XML-CHAR-0x1B][0;39;49m
```

masterleinad:
It seems like we need to update `deal.II` again such that `dealii::CUDAWrappers::get_quadrature_point<dim, double>` returns `Point<dim>`. I will also look into more device support for `dealii::Point<dim>` within `deal.II`.

Rombur:
I guess someone jumped the gun in deal.II moving from `Tensor` to `Point`...

masterleinad:
> I guess someone jumped the gun in deal.II moving from `Tensor` to `Point`...

Just for the record: the main problem is that you would expect `get_quadrature_point()` to return a `Point`-like object such that you can use it instead of a `Point`. Unfortunately, converting a `Tensor` to a `Point` is currently not possible in device code.
On the other hand, we really just need to use a more recent `deal.II` version for this pull request to work.

masterleinad:
I pushed `19.07.0` built using c7f7cca4763d84e0fb02f5ee63bccf54b802ddff.

Rombur:
Still does not pass

masterleinad:
All the `device` test are passing.

Rombur:
yeah but for that the `Device` tests to pass, we had to update deal.II and now another test breaks

masterleinad:
But that is due to updating `deal.II` and not this pull request.

masterleinad:
I just checked that the test fails for the previous commit as well.

masterleinad:
> I just checked that the test fails for the previous commit as well.

Seems like e4fe1b806f6d086218cae85c58342815c7cef55a is guilty.

masterleinad:
The corresponding pull request is https://github.com/dealii/dealii/pull/8128/files. I am really confused about how that can make any difference for the `lanczos` test (and only that test).

masterleinad:
Reverting the changes in that pull request (on top of `master`) really makes the test pass.

PR #205, opened 2019-06-21
Always test AMGX in parallel

masterleinad:
We should also run this test when the number of devices is not equal to 2.

### DIFF
--- a/tests/test_amgx_direct_solver.cu
+++ b/tests/test_amgx_direct_solver.cu
@@ -122,85 +122,81 @@ BOOST_AUTO_TEST_CASE(amgx_2_procs)
   cudaError_t cuda_error_code = cudaGetDeviceCount(&n_devices);
   mfmg::ASSERT_CUDA(cuda_error_code);
   int comm_size = dealii::Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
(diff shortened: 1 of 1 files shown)


tests/test_amgx_direct_solver.cu
```diff
@@ -122,85 +122,81 @@ BOOST_AUTO_TEST_CASE(amgx_2_procs)
   cudaError_t cuda_error_code = cudaGetDeviceCount(&n_devices);
   mfmg::ASSERT_CUDA(cuda_error_code);
   int comm_size = dealii::Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
-  if ((n_devices == 2) && (comm_size == 2))
+  if (comm_size == 2)
```
dalg24: Do we want that `if` at all then?

masterleinad:
Yes, I do.

tests/test_amgx_direct_solver.cu
dalg24: Ping

masterleinad: I don't see a reason why the test should not work with just one `MPI` process. Let's see what `CI` says in another pull request.

masterleinad: See #206.

PR #207, opened 2019-06-24
Actually call the solver in the parallel AMGX test

masterleinad:
I needed to adjust the tolerance of the outer solver to make the test pass for me locally.

### DIFF
--- a/tests/test_amgx_direct_solver.cu
+++ b/tests/test_amgx_direct_solver.cu
@@ -175,7 +175,8 @@ BOOST_AUTO_TEST_CASE(amgx_parallel)
       cusparseSetMatIndexBase(matrix_dev->descr, CUSPARSE_INDEX_BASE_ZERO);
   mfmg::ASSERT_CUSPARSE(cusparse_error_code);
   auto rhs_dev = mfmg::copy_from_host(rhs);
-  auto solution_dev = mfmg::copy_from_host(sol_ref);
+  dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::CUDA>
+      solution_dev(sol_ref.get_partitioner());
   auto params = std::make_shared<boost::property_tree::ptree>();
 
   params->put("solver.type", "amgx");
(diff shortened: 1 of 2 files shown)


tests/test_amgx_direct_solver.cu
```diff
@@ -176,7 +176,8 @@ BOOST_AUTO_TEST_CASE(amgx_2_procs)
         cusparseSetMatIndexBase(matrix_dev->descr, CUSPARSE_INDEX_BASE_ZERO);
     mfmg::ASSERT_CUSPARSE(cusparse_error_code);
     auto rhs_dev = mfmg::copy_from_host(rhs);
-    auto solution_dev = mfmg::copy_from_host(sol_ref);
+    auto solution_dev = mfmg::copy_from_host(rhs);
+    solution_dev = 0.;
```
dalg24: Was it wrong to copy `sol_ref`?  They seem to be both built with the same index range.  Was the issue initialization to zero?  I thought deal.II vector constructors where usually doing the initialization to zero, so copying from `rhs` would already have the zeroes.

tests/test_amgx_direct_solver.cu
```diff
@@ -190,6 +191,7 @@ BOOST_AUTO_TEST_CASE(amgx_2_procs)
     mfmg::CudaSolver<dealii::LinearAlgebra::distributed::Vector<
         double, dealii::MemorySpace::CUDA>>
         direct_solver_dev(cuda_handle, cuda_op, params);
+    direct_solver_dev.apply(rhs_dev, solution_dev);
```
dalg24: Weren't we getting warnings  about not using the solver?!

tests/test_amgx_direct_solver.cu
masterleinad: No, this wasn't necessarily wrong. The reasoning is just to not provide the solution as initial guess to the solver (unless we want to test for this corner case).
I first tested with `solution_dev` initialized by `rhs`, but then decided to go for zero-initialization to match the serial test case.
We could do
```
dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::CUDA>
solution_dev(sol_ref.get_partitioner());
```
instead if you prefer that.

dalg24: Sure

tests/test_amgx_direct_solver.cu
Rombur: No :(  I have seen that several times when it's a cuda file. I don't know if it's because of nvcc or if we are not passing the flags correctly.

Rombur (approved):

Rombur:
This needs to be rebased.

masterleinad:
Squashed and rebased.

PR #209, opened 2019-07-02
Add matrix-free MeshEvalutor for CUDA

Rombur:


include/mfmg/cuda/cuda_mesh_evaluator.cuh
```diff
@@ -32,9 +32,9 @@ public:
                     dealii::DoFHandler<dim> &dof_handler,
                     dealii::AffineConstraints<double> &constraints);
 
-  int get_dim() const final;
+  virtual int get_dim() const override final;
```
masterleinad: Why isn't this `static`?

include/mfmg/cuda/cuda_mesh_evaluator.cuh
```diff
@@ -32,9 +32,9 @@ public:
...
 
-  std::string get_mesh_evaluator_type() const final;
+  virtual std::string get_mesh_evaluator_type() const override;
```
masterleinad: Why isn't this `static`?

include/mfmg/cuda/cuda_mesh_evaluator.cuh
Rombur: If this function is `static`, it can't be inherited. This is possible but that means introducing more template in `hierarchy.hpp` to get ride of inheritance.

masterleinad: I see. Let's make that an issue and move on.

masterleinad: See #210.

PR #211, opened 2019-07-02
Bug fixes and small improvements

Rombur:
These are multiple bug fixes and small changes made in #204

### DIFF
--- a/include/mfmg/cuda/dealii_operator_device_helpers.cuh
+++ b/include/mfmg/cuda/dealii_operator_device_helpers.cuh
@@ -40,7 +40,7 @@ void lu_factorization(cusolverSpHandle_t cusolver_sp_handle,
                       SparseMatrixDevice<ScalarType> const &matrix,
                       ScalarType const *b, ScalarType *x);
 
-__global__ void iota(int const size, int *value, int const offset = 0);
+__global__ void iota(int const size, int *data, int const value = 0);
 } // namespace mfmg
 
 #endif
--- a/include/mfmg/cuda/dealii_operator_device_helpers.cuh
+++ b/include/mfmg/cuda/dealii_operator_device_helpers.cuh
@@ -40,7 +40,7 @@ void lu_factorization(cusolverSpHandle_t cusolver_sp_handle,
                       SparseMatrixDevice<ScalarType> const &matrix,
                       ScalarType const *b, ScalarType *x);
 
-__global__ void iota(int const size, int *value);
+__global__ void iota(int const size, int *value, int const offset = 0);
 } // namespace mfmg
 
 #endif
--- a/source/cuda/cuda_smoother.cu
+++ b/source/cuda/cuda_smoother.cu
@@ -109,11 +109,11 @@ CudaSmoother<VectorType>::CudaSmoother(
 
   ASSERT_THROW(prec_type == "jacobi", "Only Jacobi smoother is implemented.");
 
-  // Downcast the operator
-  auto cuda_operator =
+  // Downcast the operator. We need to extract the diagonal and inverse it.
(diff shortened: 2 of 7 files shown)


source/cuda/cuda_smoother.cu
```diff
@@ -109,11 +109,11 @@ CudaSmoother<VectorType>::CudaSmoother(
 
   ASSERT_THROW(prec_type == "jacobi", "Only Jacobi smoother is implemented.");
 
-  // Downcast the operator
-  auto cuda_operator =
+  // Downcast the operator. We need to extract the diagonal and inverse it.
```
masterleinad:
```suggestion
  // Downcast the operator. We need to extract the diagonal and invert it.
```

source/cuda/cuda_smoother.cu
```diff
@@ -132,8 +132,8 @@ CudaSmoother<VectorType>::CudaSmoother(
   int *row_index_coo_dev = nullptr;
   cuda_malloc(row_index_coo_dev, local_nnz);
 
-  // Change to COO format. The only thing that needs to be change to go from CSR
-  // to COO is to change row_ptr_dev with row_index_coo_dev.
+  // Change to COO format. The only thing that needs to be change to go from
```
masterleinad:
```suggestion
  // Change to COO format. The only thing that needs to be changed to go from
```

source/cuda/cuda_solver.cu
```diff
@@ -228,8 +228,7 @@ CudaSolver<VectorType>::CudaSolver(
 
     // We need to cast away the const because we need to change the format
     // of _matrix to the format supported by amgx
-    auto &amgx_matrix = const_cast<SparseMatrixDevice<value_type> &>(
-        *(cuda_operator->get_matrix()));
+    auto &amgx_matrix = *(cuda_operator->get_matrix());
```
masterleinad: This is really confusing. In the end, `matrix` is changed heavily here. For me, it would make more sense if the constructor copies the object.

include/mfmg/cuda/dealii_operator_device_helpers.cuh
```diff
@@ -40,7 +40,7 @@ void lu_factorization(cusolverSpHandle_t cusolver_sp_handle,
                       SparseMatrixDevice<ScalarType> const &matrix,
                       ScalarType const *b, ScalarType *x);
 
-__global__ void iota(int const size, int *value);
+__global__ void iota(int const size, int *value, int const offset = 0);
```
dalg24: I find the naming of the variables disturbing:
* `offset` usually refers to a distance between memory location.
* in all the documentations of `std::iota` I have seen, the last argument that specifies the initial value is always called `val` or `value`.

I suggest you rename:
* `value` -> `first` or `data`
* `offset` -> `value`

source/cuda/cuda_solver.cu
Rombur: I don't understand what's the problem here.

masterleinad: Passing `op` as `std::shared_ptr` to a `const` object feels conceptually weird if we change the wrapped object.

Rombur: I see but 
  1) I will not copy the object. This is the most expensive part memory-wise of the algorithm. 
  2) We do have a few variables (include the data in `SparseMatrixDevice`) that are `mutable` and while we shouldn't use `mutable` this is out of the scope of this PR.

Rombur:
I have updated the PR

source/cuda/cuda_solver.cu
masterleinad: Let's document these problems and move on.

PR #213, opened 2019-07-10
Add CUDA matrix-free operator

Rombur:
This is almost identical to the matrix-free operator on the host.

### DIFF
--- a/include/mfmg/cuda/cuda_matrix_free_operator.cuh
+++ b/include/mfmg/cuda/cuda_matrix_free_operator.cuh
@@ -33,13 +33,18 @@ public:
 
   void vmult(vector_type &dst, vector_type const &src) const;
 
-  // Needed because the extract_row works only on the host
+  // Needed because the extract_row() works only on the host. Thus this function
+  // always needs to exist.
   void apply(dealii::LinearAlgebra::distributed::Vector<
                  value_type, dealii::MemorySpace::Host> const &x,
              dealii::LinearAlgebra::distributed::Vector<
                  value_type, dealii::MemorySpace::Host> &y,
              OperatorMode mode = OperatorMode::NO_TRANS) const;
 
+  // While this function is templated on the vector_type, the vector_type is
+  // always, dealii::LinearAlgebra::distributer::Vector<value_type,
+  // dealii::MemorySpace::CUDA> It is templated on the vector_type for
+  // historical reason.
   virtual void apply(vector_type const &x, vector_type &y,
                      OperatorMode mode = OperatorMode::NO_TRANS) const override;
 
--- a/include/mfmg/cuda/cuda_matrix_free_operator.cuh
+++ b/include/mfmg/cuda/cuda_matrix_free_operator.cuh
@@ -1,0 +1,75 @@
+/*************************************************************************
+ * Copyright (c) 2017-2019 by the mfmg authors                           *
+ * All rights reserved.                                                  *
+ *                                                                       *
+ * This file is part of the mfmg libary. mfmg is distributed under a BSD *
+ * 3-clause license. For the licensing terms see the LICENSE file in the *
+ * top-level directory                                                   *
+ *                                                                       *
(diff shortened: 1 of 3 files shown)


source/cuda/cuda_matrix_free_operator.cu
```diff
@@ -0,0 +1,188 @@
...
+  // FIXME Return garbage since throwing not implemented will make important
+  // tests to fail
+  return 0;
+}
+
+template <int dim, typename VectorType>
+size_t CudaMatrixFreeOperator<dim, VectorType>::operator_complexity() const
+{
+  // FIXME Return garbage since throwing not implemented will make important
+  // tests to fail
```
masterleinad: Which test would fail?

Rombur: `test_hierarchy_device` I guess we can also remove the line in that output the grid and the operator complexity like we did on the host where we hit the same problem.

source/cuda/cuda_matrix_free_operator.cu
```diff
@@ -0,0 +1,188 @@
...
+  ASSERT_CUSPARSE(cusparse_error_code);
+  cusparse_error_code =
+      cusparseSetMatType(c_dev->descr, CUSPARSE_MATRIX_TYPE_GENERAL);
+  ASSERT_CUSPARSE(cusparse_error_code);
+  cusparse_error_code =
+      cusparseSetMatIndexBase(c_dev->descr, CUSPARSE_INDEX_BASE_ZERO);
+  ASSERT_CUSPARSE(cusparse_error_code);
+
+  std::shared_ptr<Operator<VectorType>> op(
+      new CudaMatrixOperator<VectorType>(c_dev));
```
masterleinad: You should be able to just do
```
return std::make_shared<CudaMatrixOperator<VectorType>>(c_dev);
``` 
here.

source/cuda/cuda_matrix_free_operator.cu
```diff
@@ -0,0 +1,188 @@
...
+}
+
+template <int dim, typename VectorType>
+std::shared_ptr<VectorType>
+CudaMatrixFreeOperator<dim, VectorType>::build_domain_vector() const
+{
+  // We know the operator is squared
+  auto domain_vector = this->build_range_vector();
+
+  return domain_vector;
```
masterleinad: You can also just compress this to
```
return this->build_range_vector();
```

include/mfmg/cuda/cuda_matrix_free_operator.cuh
```diff
@@ -0,0 +1,75 @@
...
+
+  // Needed because the extract_row works only on the host
+  void apply(dealii::LinearAlgebra::distributed::Vector<
+                 value_type, dealii::MemorySpace::Host> const &x,
+             dealii::LinearAlgebra::distributed::Vector<
+                 value_type, dealii::MemorySpace::Host> &y,
+             OperatorMode mode = OperatorMode::NO_TRANS) const;
+
+  virtual void apply(vector_type const &x, vector_type &y,
+                     OperatorMode mode = OperatorMode::NO_TRANS) const override;
```
dalg24: I am getting confused but the coexistence of `vmult()` and `apply()` overload that take `vector_type` and the `apply()` overload that takes vectors on the host.  Was the latter introduced so you can reuse `matrix_transpose_matrix_multiply()`?

Rombur: Yes exactly. I'll add a comment.

source/cuda/cuda_matrix_free_operator.cu
Rombur: meh that's just a stylistic difference

Rombur:
I have added a comment about the different `apply()`. I haven't done anything for `operator_complexity()` and `grid_complexity()` because they will disappear in the future.

Rombur:
_ping_

Rombur:
_ping_

PR #215, opened 2019-07-12
Link against non-debug deal.II for RelWithDebInfo build

aprokop:


cmake/SetupTPLs.cmake
```diff
@@ -48,7 +48,7 @@ ENDIF()
 # enable dealii::Assert in header files.
 STRING(FIND "${DEAL_II_LIBRARIES}" "general" SINGLE_DEAL_II)
 IF (${SINGLE_DEAL_II} EQUAL -1)
-  IF(CMAKE_BUILD_TYPE MATCHES "Release")
+  IF(CMAKE_BUILD_TYPE MATCHES "Release" OR CMAKE_BUILD_TYPE MATCHES "RelWithDebInfo")
```
masterleinad: What about `MinSizeRel`? :wink:

PR #216, opened 2019-07-16
Reuse eigenvectors as initial guess for LOBPCG

Rombur:
The basic idea here is to use `ScratchData` to update the initial guess for LOBPCG. `ScratchData` is local to a thread which means that using a different number of threads will change the initial guess for a given agglomerate. Other than that, I had to change the constructor of `mfmg::MultiVector` to use a `std::vector` instead of a `VectorType`. In my test (in debug mode), the setup was 15 to 20% faster.

### DIFF
--- a/include/mfmg/dealii/amge_host.hpp
+++ b/include/mfmg/dealii/amge_host.hpp
@@ -17,6 +17,14 @@
 
 namespace mfmg
 {
+/**
+ * Use the structure to store initial guess for LOBPCG
+ */
+struct LobpcgScratchData
+{
+  std::vector<dealii::Vector<double>> lobpcg_init_guess;
+};
+
 template <int dim, typename MeshEvaluator, typename VectorType>
 class AMGe_host : public AMGe<dim, VectorType>
 {
@@ -37,6 +45,7 @@ public:
    *  - an object that can evaluates the local DoFHandler, the local
    *    AffineConstraints<double>, and the local system sparse matrix with its
    *    sparsity pattern.
+   *  - an object that contains initial guesses for LOBPCG
    *
    * The function returns the complex eigenvalues, the associated eigenvectors,
    * the diagonal elements of the local system matrix, and a vector that maps
@@ -60,7 +69,7 @@ public:
       std::map<typename dealii::Triangulation<dim>::active_cell_iterator,
                typename dealii::DoFHandler<dim>::active_cell_iterator> const
           &patch_to_global_map,
-      MeshEvaluator const &evaluator,
+      MeshEvaluator const &evaluator, LobpcgScratchData const &scratch_data,
       typename std::enable_if_t<is_matrix_free<MeshEvaluator>::value &&
                                     std::is_class<Triangulation>::value,
                                 int> = 0) const;
@@ -75,7 +84,7 @@ public:
       std::map<typename dealii::Triangulation<dim>::active_cell_iterator,
                typename dealii::DoFHandler<dim>::active_cell_iterator> const
           &patch_to_global_map,
-      MeshEvaluator const &evaluator,
+      MeshEvaluator const &evaluator, LobpcgScratchData const &scratch_data,
       typename std::enable_if_t<!is_matrix_free<MeshEvaluator>::value &&
                                     std::is_class<Triangulation>::value,
                                 int> = 0) const;
@@ -107,14 +116,6 @@ public:
 
 private:
   /**
-   * This data structure is empty but it is necessary to use WorkStream.
-   */
-  struct ScratchData
-  {
-    // nothing
-  };
-
-  /**
    * Structure which encapsulates the data that needs to be copied at the end
    * of WorkStream.
    */
@@ -133,7 +134,7 @@ private:
   void local_worker(unsigned int const n_eigenvectors, double const tolerance,
                     MeshEvaluator const &evaluate,
                     std::vector<unsigned int>::iterator const &agg_id,
-                    ScratchData &scratch_data, CopyData &copy_data);
+                    LobpcgScratchData &scratch_data, CopyData &copy_data);
 
   /**
    * This function copies quantities computed in local worker to output
--- a/include/mfmg/dealii/amge_host.templates.hpp
+++ b/include/mfmg/dealii/amge_host.templates.hpp
@@ -208,25 +208,64 @@ void lanczos_compute_eigenvalues_and_eigenvectors(
 
 template <typename AgglomerateOperator>
 void anasazi_compute_eigenvalues_and_eigenvectors(
-    unsigned int n_eigenvectors, double tolerance,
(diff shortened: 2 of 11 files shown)


masterleinad:
Hmm... Testing this branch all the results are slower and require more iterations.

Rombur:
So apparently the problem is that in 3D the tolerance of `1e-2` is not tight enough. @masterleinad is trying to figure out what it should be but `1e-3` already gives results that are more sane.

aprokop:
So, does it mean that we need to compute the eigenvectors for reuse with higher precision? I don't think the tolerance for aggregates that use the reused initial guess should be tightened.

I don't really understand how reusing the vectors increases the number of iterations. It would mean that it performs worse the random guess, which is weird.

Rombur:
When @masterleinad talks about iterations, he is talking about the number of CG iterations not the number of LOBPCG iterations. In 3D, when we don't use the initial guess changing the LOBPCG tolerance from `1e-2` to `1e-3` divides the number of CG iterations by 3.

The tolerance seems to work fine in 2D but not for 3D. With the tolerance at `1e-2`, in 2D we need about 30 iterations to get the eigenvectors but in 3D, LOBPCG converges in 3 (!!) iterations. If we tighten this to `1e-3` it takes about 30 iterations to converge and the number of CG is smaller and more stable.

masterleinad:
Using a higher smoother polynomial degree the difference in the number of CG iterations is not as large, but rather like 13 vs 25 iterations. I am still surprised to see different numbers at all.
Anyway, for that particular case, I see 9 vs 9 iterations for 1.e-3, 12 vs. 10 iterations for 2.e-3 and 16 vs. 10 iterations for 5.e-3.

In the end, we just have to decide for one setting we want to produce results with. 1.e-3 seems to be a good choice and is for this one case about 10% more expensive.

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -557,9 +600,18 @@ void AMGe_host<dim, MeshEvaluator, VectorType>::local_worker(
...
-      compute_local_eigenvectors(n_eigenvectors, tolerance,
-                                 agglomerate_triangulation,
-                                 agglomerate_to_global_tria_map, evaluator);
+      compute_local_eigenvectors(
+          n_eigenvectors, tolerance, agglomerate_triangulation,
+          agglomerate_to_global_tria_map, evaluator, scratch_data);
+
+  if (_eigensolver_params.get("type", "lanczos") == "anasazi")
+  {
+    if (_eigensolver_params.get("use_initial_guess", false))
```
masterleinad: This parameter should be visible in `hierarchy_input.info`.

include/mfmg/dealii/anasazi.templates.hpp
```diff
@@ -78,6 +79,12 @@ AnasaziSolver<OperatorType, VectorType>::solve(
   Anasazi::Eigensolution<double, MultiVectorType> solution =
       solver->getProblem().getSolution();
 
+  int const verbosity = params.get("verbosity", 0);
```
masterleinad: Again, I think it's useful to see this option in `hierarchy_input.info`.

tests/hierarchy_driver.cc
```diff
@@ -269,6 +269,7 @@ int main(int argc, char *argv[])
   int const fe_degree = params->get<unsigned int>("laplace.fe_degree", 4);
   params->put("fast_ap", true);
   params->put("eigensolver.type", "anasazi");
+  params->put("eigensolver.tolerance", 1e-2);
```
masterleinad: What do we want here finally? `1.e-3`?

tests/test_hierarchy.cc
```diff
@@ -372,6 +382,27 @@ BOOST_DATA_TEST_CASE(
     ref_solution[std::make_tuple("hyper_ball" , "distort"    , "None"                  , "anasazi" , "matrix-free")] = 0.3029292535;
     ref_solution[std::make_tuple("hyper_ball" , "distort"    , "Reverse Cuthill_McKee" , "anasazi" , "matrix-full")] = 0.0970940159;
     ref_solution[std::make_tuple("hyper_ball" , "distort"    , "Reverse Cuthill_McKee" , "anasazi" , "matrix-free")] = 0.3029292535;
+
+    // FIXME Running ./test_hierarchy instead of ctest -R test_hierarchy gives a
+    // better convergence rate. Tightening the tolerance on anasazi-does reduces
```
masterleinad: Why would `./test_hierarchy` and `ctest -R test_hierarchy` give different results? Don't they run the exact same executable with the same input data?

include/mfmg/dealii/amge_host.templates.hpp
```diff
@@ -208,25 +208,64 @@ void lanczos_compute_eigenvalues_and_eigenvectors(
...
+          {
+            (*lobpcg_initial_guess[j])[i] = initial_guess[i];
+          }
+        }
+      }
+    }
+  }
   std::tie(real_eigenvalues, eigenvectors) =
-      solver.solve(anasazi_params, initial_guess);
+      solver.solve(eigensolver_params, lobpcg_initial_guess);
```
masterleinad: So we update lobpcg_init_guess also in this call?

include/mfmg/dealii/amge_host.hpp
```diff
@@ -17,6 +17,14 @@
 
 namespace mfmg
 {
+/**
+ * Use the structure to store initial guess for LOBPCG
+ */
+struct LobpcgScratchData
```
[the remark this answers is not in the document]
Rombur: Yeah that's how I wrote it at first but I found that it was harder to read.

include/mfmg/dealii/amge_host.templates.hpp
Rombur: What do you mean? Here we just call the solver and use whatever we have set as initial guess.

include/mfmg/dealii/anasazi.templates.hpp
Rombur: Sure I can add these two in `hierarchy_input.info` file.

tests/hierarchy_driver.cc
Rombur: Yeah I think that makes sense to change it.

tests/test_hierarchy.cc
Rombur: Yeah I have no clue what's the difference.

include/mfmg/dealii/amge_host.templates.hpp
masterleinad: It's https://github.com/ORNL-CEES/mfmg/pull/216/files#diff-5896bf281196f7deec942b95d48b0a90R612

Rombur:
_ping_

PR #217, opened 2019-07-29
Add CUDA matrix-free smoother

Rombur:


source/cuda/cuda_matrix_free_smoother.cu
```diff
@@ -0,0 +1,148 @@
...
+    std::shared_ptr<Operator<VectorType> const> op,
+    std::shared_ptr<boost::property_tree::ptree const> params)
+    : Smoother<VectorType>(op, params)
+{
+  std::string prec_type = this->_params->get("smoother.type", "Jacobi");
+  // Transform to lower case
+  std::transform(prec_type.begin(), prec_type.end(), prec_type.begin(),
+                 tolower);
+
+  ASSERT_THROW(prec_type == "jacobi", "Only Jacobi smoother is implemented.");
```
masterleinad, line 88: We should soon be able to use `PreconditionChebyshev` as well.

source/cuda/cuda_matrix_free_smoother.cu
```diff
@@ -0,0 +1,148 @@
...
+
+  // x = x + B^{-1} (-r)
+  dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::CUDA>
+      tmp(x);
+  smoother.vmult(tmp, r);
+  x.add(-1., tmp);
+}
+
+template <typename ScalarType>
+__global__ void extract_inv_diag(ScalarType const *const matrix_value,
```
masterleinad, line 65: Looks like this is never used, is it?

source/cuda/cuda_matrix_free_smoother.cu
```diff
@@ -0,0 +1,148 @@
...
+        VectorType const &b, VectorType &x);
+};
+
+template <typename VectorType>
+void SmootherOperator<VectorType>::apply(
+    Operator<VectorType> const &op,
+    SparseMatrixDevice<typename VectorType::value_type> const &smoother,
+    VectorType const &b, VectorType &x)
+{
+  ASSERT_THROW_NOT_IMPLEMENTED();
```
masterleinad, line 37: Do we actually need to instantiate the operator for any type different from `dealii::LinearAlgebra::distributed::Vector<
    double, dealii::MemorySpace::CUDA>>`? If not, we could just use a `static_assert` instead.

source/cuda/cuda_matrix_free_smoother.cu
```diff
@@ -0,0 +1,148 @@
...
+  dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::CUDA>
+      r(b);
+  op.apply(x, r);
+  r.add(-1., b);
+
+  // x = x + B^{-1} (-r)
+  dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::CUDA>
+      tmp(x);
+  smoother.vmult(tmp, r);
+  x.add(-1., tmp);
```
masterleinad, line 61: There is nothing special about `dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::CUDA>` here, is it? If we can indeed use `static_assert`, this would look much nicer with `VectorType` instead.

source/cuda/cuda_matrix_free_smoother.cu
```diff
@@ -0,0 +1,148 @@
...
+#include <mfmg/cuda/cuda_matrix_operator.cuh>
+#include <mfmg/cuda/dealii_operator_device_helpers.cuh>
+#include <mfmg/cuda/utils.cuh>
+
+namespace mfmg
+{
+namespace
+{
+template <typename VectorType>
+struct SmootherOperator
```
masterleinad, line 23: So all this is just for being able to specialize for `VectorType`. Do we actually need that if only allow for one type?

PR #218, opened 2019-10-04
Don't use eigenvectors as initial guess for constrained agglomerates

Rombur:
@masterleinad @aprokop With this PR, we don't reuse the eigenvectors as initial guess for the eigensolver if the agglomerate is on the boundary. If the agglomerate is inside the domain but the initial guess is from a boundary eigenvectors, we add the non-zero entries. This particular implementation only works for matrix-free so I couldn't really test it but I tested something very close that worked with matrix. The good news is that the convergence rate matches the convergence rate when we are not using an initial guess. So I would expect that the difference in the number of CG iterations disappear. However, in the test the meshes are very small so a very large number of agglomerates are on the boundary. Thus, there are very few agglomerates that actually use the initial guess. I can create a version that works also on matrix-based problems. Basically the difference is that we don't have access to the `constraints` to know if the agglomerate is on the boundary. In that case, we need to go through the initial guess vector and check if an entry has been set to zero.

### DIFF
--- a/include/mfmg/dealii/amge_host.templates.hpp
+++ b/include/mfmg/dealii/amge_host.templates.hpp
@@ -98,7 +98,7 @@ struct MatrixFreeAgglomerateOperator
    */
   size_type n() const { return _dof_handler.n_dofs(); }
 
-private:
+  // private:
   /**
    * The actual operator wrapped.
    */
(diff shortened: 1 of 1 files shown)


aprokop:
We are not interested in matrix-based version, I think.

> If the agglomerate is inside the domain but the initial guess is from a boundary eigenvectors, we add the non-zero entries.

This is different from what I expected. I thought we have no cross-pollination between boundary and internal agglomerates at all. Could you remind how you insert the nonzero entries?

masterleinad:
The difference decreases but doesn't vanish...

